PR #175, opened 2014-09-23
WIP - Adding automation scripts and files for l4d2

bfosberry:
- Adding Dockerfile to set up a container
- Adding Vagrantfile to build the image locally
- Adding l4d2server_automated script bypassing reads

DO NOT MERGE

This is an example of how to containerize one of the game servers. 

1) Create an automatable version of the l4d2server manager script. This could be the main script with an expect script instead
2) Set up a Dockerfile to prep an ubuntu box, download the script, and run the install.
3) Write a custom server.cfg over the default one
4) (not yet implemented) set up an entrypoint to start the server, wrapped behind the monitor
5) (not yet implemented) expose the ports in the container
6) Wrap the container in DNS so that it can report its hostname correctly

Honestly this is a bit of a departure from where I need to be, I think a lot of these tools are useful in a non-automated environment, but I may stick to the bare essentials and use the base commands in my setup

### DIFF
--- a/Left4Dead2/Dockerfile
+++ b/Left4Dead2/Dockerfile
@@ -1,0 +1,23 @@
+# Left4Dead2 Server repo
+#
+# VERSION 0.1
+
+FROM ubuntu
+MAINTAINER bfosberry
+
+RUN apt-get update
+ENV DEBIAN_FRONTEND noninteractive
+RUN apt-get -y install gdb mailutils postfix lib32gcc1 expect wget
+
+ENV USERNAME l4d2server
+RUN adduser --disabled-password --gecos "" $USERNAME
+RUN usermod -a -G sudo $USERNAME
+ENV HOME /home/$USERNAME
+
+ADD ./l4d2server_automated /home/l4d2server/
+RUN chmod +x /home/l4d2server/l4d2server_automated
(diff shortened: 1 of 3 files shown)


zipper97412:
I am intresting in puting all of the servers in a docker container, i managed to create a csgo server using the original script and i put the bash terminal as 1st Prosses to force the container to stay open i also made a litle bash script outside to run and/or attach the terminal of the container but it is VERY ugly, unpractical and it dosent folow the docker philosophy but it work... i think it is possible to improve the system (find an other 1st prosses to manage that shit) the advantage is that you dont have to rewrite the server's script... if you have an idea?

bfosberry:
Fyi Im working on a separate spec for docker game servers: http://bfosberry.github.io/gscs/

I think it would be awesome to be able to split these scripts up and use them in the relevant container sections

johnjelinek:
:laughing: I just found this. After making these:
[CS:GO Server - Docker](https://hub.docker.com/r/johnjelinek/csgoserver/)
[L4D2 Server - Docker](https://hub.docker.com/r/johnjelinek/l4d2server/)
[TF2 Server - Docker](https://hub.docker.com/r/johnjelinek/tf2server/)

PR #276, opened 2015-01-25
Fix : binary operator expected and don't use a not not operator

Poil:


Poil:
Update: True selfname. Remove stupid code

dgibbs64:
Thanks for the code updates. Looks like you know best practices better than me. I welcome any code that improves or optimises the scripts. I will review your code and do a merge very soon. Thanks for submitting this.

Poil:
Hi, perhaps we can also move all "Directories, Logging, fn_runfunction" section from script to a common shared configuration file and just source this file.

Poil:
Also you have some duplicate code (ex: fn_distro, fn_distro_detail)
You can use "/proc/loadavg"  and "/proc/meminfo" to replace uptime/free command with a simple awk

And for the best practices, In Bash either you use simple bracket syntax and you protect your variable with double quotes or you use the double bracket syntax
```bash
[ -z "${foo}" ]
#or
[[ -z ${foo} ]]
#or ...
[[ -z "${foo}" ]]
```

You can have more syntax improvment on Bash hackers wiki http://wiki.bash-hackers.org/start
Note, you can replace all your if or syntax : http://wiki.bash-hackers.org/syntax/ccmd/conditional_expression

dgibbs64:
I don't quite follow. fn_runfunctions is required to get all other functions. I'm not sure of a way to move the function out of the main file.

PR #292, opened 2015-02-14
Teamspeak3 - Wrong Name of Lock file 

gOOvER:
For TS3 it must be: ts3server.pid - not .ts3server.lock


dgibbs64:
this is used for the script and not for the teamspeak server itself no need to change this :)

Issue #274, opened 2015-01-25
[TeamFortressClassic] Stuck in update loop


Xylemon:
On other games like Killing Floor, updating is fine and reports the following:

[  OK  ] Update kf-server: Checking for update: SteamCMD

No update available:
	Current version: 460851
	Available version: 460851
	https://steamdb.info/app/215360/

But when done with TFC, this happens:

[ WARN ] Update tfc-server: No appmanifest_90 +app_set_config 90 mod tfc.acf fou
[ INFO ] Update tfc-server: Forcing update to correct issue
[ .... ] Update tfc-server: Updating Chuck Bronson's Prestige and Well-Mannered 
[  OK  ] Update tfc-server: Updating Chuck Bronson's Prestige and Well-Mannered Debate Lounge [BETA]

And keeps updating over and over until the the server admin stops the script.

Scarsz:
OS/arch

dgibbs64:
Interesting issue. It probably has something to do with how gold source games handling app manifest files differently. I will look in to a fix for this.....Gold source games are a big pain in my back side most of the time.

dgibbs64:
I have been unable to download the server files again. I have updated this issue https://github.com/ValveSoftware/halflife/issues/1583#issuecomment-73278666 however its not promising.

Xylemon:
Considering Valve still has broken custom sprays in GoldSrc games, I too am doubtful they'll fix the issue. Thanks for trying to get the source of this problem.

dgibbs64:
Im planning on bypassing the update checker for goldsrc games. SteamCMD is still pretty broken for goldsrc. at least you will be able to run the update-restart feature for these servers. Stay tuned for updates

dgibbs64:
merging with issue #323

PR #314, opened 2015-03-16
Added scripts for the rest of the GoldSrc Steam games

### LINKED ISSUE #240: [HL:Opposing Force/Ricochet] Creation
Could support be added for the rest of the Steam GoldSrc Engine based games? Like Opposing Force, HLDM, and Ricochet (well if you really want to do that one).

Xylemon:
This should fix #240 however, I did have problems getting the mods to actually download. Perhaps the problem is related to #274.

### DIFF
--- a/DeathmatchClassic/cfg/lgsm-default.cfg
+++ b/DeathmatchClassic/cfg/lgsm-default.cfg
@@ -1,0 +1,35 @@
+// Server Name
+hostname "hostname"
+
+// RCON Password
+rcon_password "rconpassword"
+
+// Server Password
+sv_password ""
+
+// Server Logging
+log on
+sv_logbans 1
+sv_logecho 1
+sv_logfile 1
+sv_log_onefile 0
+
+// disable autoaim
+sv_aim 0
+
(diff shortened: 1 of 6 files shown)


dgibbs64:
Great work. Just a question about HL:DMC I already have a server called Half Life: Deathmatch classic. mine says valve and yours dmc. Whats the difference?

Xylemon:
The one you have with the "valve" folder is actually regular Half-Life multiplayer. Deathmatch Classic or "dmc" is a port of Quake by Valve to GoldSrc. You should change the HalfLifeDeathmatchClassic folder to just HalfLifeDeathmatch.

dgibbs64:
Thanks for the heads up I will modify my scripts and merge yours. Thanks for your submission. Code submissions are always welcome.

Xylemon:
No problem, I'll probably contribute some more soon enough here.

dgibbs64:
Looks like we are only missing Half-Life Deathmatch: Source Dedicated server now and we have the complete set. If you could create that one it would be very helpful.

PR #326, opened 2015-03-30
gamemode before map

darkjacky:
It is possible to have maps in the gamemodes folder in that case its a requirement for the game to know the gamemode is before it can change to a map.

If srcds can not find the map it will refuse to start.


dgibbs64:
I don't really understand why switching two parms around makes a different. I woudl of thought that if the map wasn't found it would fail anyway

PR #328, opened 2015-04-06
fix: TS3 Version check

MAFLO321:
Now Versions are sorted against Name.

Bug: The update was installing v3.0.3 over my v3.0.11.2!
```
./ts3server update
[  OK  ] Update ts3-server: Checking for update: teamspeak.com

Update available:
        Current build: 3.0.11.2
        Available build: 3.0.3


Applying update...
[...]
```


dgibbs64:
Thanks for the fix :D

Issue #318, opened 2015-03-17
L4D2server script fails to find lock file


patriotgames:
<h4>Background</h4>

Testing your latest L4D2server script on two different VMs, one with a fresh install of **Ubuntu Server 14.04.2 LTS** and the other with a fresh install of **Debian 8 Jessie amd64.**  All required 32-bit libraries and other packages were installed per your instructions at:

http://danielgibbs.co.uk/lgsm/l4d2server/

L4D2server script completes the Initial game server installation without issue.  Script functions (start, stop, restart, update, validate, etc.) all work correctly.

Automated monitoring setup by downloading `gsquery.py` (by script during initial installation) and a new cron job created for 3 minute checks:

`*/3 * * * * /home/patriot/gameservers/l4d2server monitor > /dev/null 2>&1`

Issuing the command `./l4d2server start` the game server starts correctly without issue and a `.l4d2-server.lock` file is created in the same directory where the script and `gsquery.py` exist.

<h4>Issue</h4>

The `l4d2-server-script.log` located in `/home/patriot/gameservers/log/script` has these entries:

Mar 17 13:21:01 l4d2-server: Monitor: ubuntu test
Mar 17 13:21:02 l4d2-server: Monitor: Disabled: No lock file found
Mar 17 13:24:01 l4d2-server: Monitor: ubuntu test
Mar 17 13:24:02 l4d2-server: Monitor: Disabled: No lock file found

and yet **the lock file clearly exists.**  Debian 8 produces the same results in its script log.

<h4>Result</h4>

Because the script fails to find the `.l4d2-server.lock` file, **the game server monitoring process does not work** and if the game server is forced to crash with the rcon command `quit`, **the server is not restarted by the cron job.**

This is the only issue I've found with your latest script.  If you are able to resolve it the script seems very solid.

Thank you for all of your hard work developing these scripts and for making them available to the community free of charge.

Cheers,
Patriot

dgibbs64:
Hey. Can I have the output of details please and also ls -al of the main script directory.

patriotgames:
Hi,

 

Of course.

 

pwd

/home/patriot/gameservers

 

ls -a

.  ..  functions  gsquery.py  l4d2server  .l4d2-server.lock  log  serverfiles  steamcmd

 

[...]

PR #403, opened 2015-05-23
Fixed monitoring for cron job

xFaNaTiix:
This should fix "monitor" when it's run by a cron job or something like this. (e.g. #318)
Please take a look at it. It works for me. =)

### DIFF
--- a/functions/fn_monitor
+++ b/functions/fn_monitor
@@ -16,7 +16,7 @@ fn_logs
 fn_printdots "${servername}"
 fn_scriptlog "${servername}"
 sleep 1
-if [ ! -f "${lockselfname}" ]; then
+if [ ! -f "${rootdir}/${lockselfname}" ]; then
 	fn_printinfo "Disabled: No lock file found"
 	fn_scriptlog "Disabled: No lock file found"
 	sleep 1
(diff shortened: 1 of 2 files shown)


dgibbs64:
Interesting didn't realise there was a problem. I run monitor and haven't had an issue before. What distro you running? Also any ideas why this is an issue for you? Im happy to merge once i have tested but im curious on why this is an issue.

xFaNaTiix:
Distro Details:
Distro:    Ubuntu 14.04.2 LTS
Arch:      x86_64
Kernel:    3.13.0-35-generic
Hostname:  puck372
tmux:      tmux 1.8
GLIBC:     2.19


*/5 * * * * /home/marvin/gameservers/csgoserver monitor > /dev/null 2>&1

If I run "./csgoserver monitor" it works fine but If this command is run by a cron job:
Mai 23 17:26:02 csgo-server: Monitor: Disabled: No lock file found

Apparently there are path issues. So I changed the path to the full path. https://github.com/xFaNaTiix/linuxgsm/blob/6bb595cc33d053e5f25966a91d57c68b10d384fb/functions/fn_monitor#L64

Then the first issue was gone but then:
Mai 23 17:27:03 csgo-server: Monitor: Checking session: CHECKING
Mai 23 17:27:04 csgo-server: Monitor: Checking session: OK

..it didn't start gsquery.py because it didn't find it.

So I changed all paths to the full paths and it worked (fn_monitor_query) :)

Scarsz:
Commits look good and even if it's not an issue for some people it's in good practice to have full pathnames to files.

Issue #395, opened 2015-05-19
Add option to suppress console warning


athairus:
After launching the console dozens of times, I think these lessons are burned into my head:
- I should not press CTRL+C
- I should exit by pressing CTRL+B, D

I'd be nice if I didn't have to type `y, Enter` every time I want to open the console up.

Alternatively, just have Y be the default option so I can just press enter to continue.

dgibbs64:
added the default y in a while ago.

Issue #311, opened 2015-03-06
[l4d1Server] LAN servers are restricted to local clients (class C)


faflfama:
Hi,
I just set up a server, which then Starts fine. When I try to connect it shows "LAN servers are restricted to local clients (class C)". I tried to set sv_lan "0", but that didnt help. Did I do something wrong?

Fabian D.

faflfama:
No solution?

dgibbs64:
Not sure on this one. I will investigate when I can. However there may be solutions around as this isn't a script issue.

faflfama:
Thanks for the answer, alt least Im glad I didnt do sth wrong.

faflfama:
Solved it for me:
Copied a steamclient.so from a working L4D2 and copied it to "steamcmd/linux32" and ".steam/sdk32". After that the server started normally.

nikolaydan:
This bug is depending of OS, I tried different Linux distribs and server worked with Ubuntu 14.10 and CentOS 7.1.
I had this error with Ubuntu 14.04, Centos 6, Centos 7.0, Debian 7.

I used EXACTLY the same files (i used .tar archive and tried same archive on different OS).

Issue #230, opened 2014-12-08
[l4dserver] - Unknown command "servercfgfile"


kdavidb:
L4D server installed and controlled via the script does not run properly.  Ubuntu 10.04 leased VM that does run other low-utilization, non-gaming servers (web and videoconferencing).  

L4D server installation per the website (including dependencies) completes without report of error other than  "ERROR: certificate common name `www.github.com' doesn't match requested host name `raw.githubusercontent.com'." which is bypassed with the suggested '--no-check-certificate' flag.  This installation was repeated three times and the results were identical.  

The server does start and appears to run, but is limited as a LAN-only server and refuses remote clients.  After starting, the console holds this information:
-*-------------------------------------------------------------------*-
Server will auto-restart if there is a crash.
Running a benchmark to measure system clock frequency...
Finished RDTSC test. To prevent the startup delay from this benchmark, set the environment variable RDTSC_FREQUENCY to 2327.000000 on this system. This value is dependent upon the CPU clock speed and architecture and should be determined separately for each server. The use of this mechanism for timing can be disabled by setting RDTSC_FREQUENCY to 'disabled'.
#Using breakpad minidump system
Using breakpad crash handler
#
#Console initialized.
Unable to remove /home/l4dserver/serverfiles/left4dead/addonlist.txt!
#Unable to remove /home/l4dserver/serverfiles/left4dead/addonlist.txt!
#Game.dll loaded for "L4D - Co-op - Normal"
#ConVarRef cl_language doesn't point to an existing ConVar
Server is hibernating
Queuing heartbeat send to legacy master server.
Game supporting (2) split screen players
maxplayers set to 18
Unable to remove /home/l4dserver/serverfiles/left4dead/addonlist.txt!
Unable to remove /home/l4dserver/serverfiles/left4dead/addonlist.txt!
Unknown command "servercfgfile"
Network: IP XXX.XXX.XXX.XXX, mode MP, dedicated Yes, ports 27015 SV / 0 CL
---- Host_NewGame ----
Host_NewGame on map l4d_hospital01_apartment
ConVarRef room_type doesn't point to an existing ConVar
Executing dedicated server config file
Commentary: Loading commentary data from maps/l4d_hospital01_apartment_commentar
y.txt.
[S_API FAIL] SteamAPI_Init() failed; unable to locate a running instance of Steam, or a local steamclient.dll.
-************************************************
*-  Unable to load Steam support library.       -*
[...]

PR #406, opened 2015-05-26
A few fixes and additions

xFaNaTiix:
**Fixed 'Unknown command "servercfgfile"'** ( #230 )
Just removed it because it doesn't exists in L4D1.

**Fixed Steam / LAN error** ( #230 | #311 )
This fixes the error message in console but I'm not sure about the ingame problem because I don't have this game to check it.

**Added default input "y" when opening console** ( #395 )
I think this is the easiest way to implement it but it requires bash v4. I don't know if it is a problem.

**Added lock file removal for all servers**
I think it wasn't intended to be missing?

### DIFF
--- a/Left4Dead/l4dserver
+++ b/Left4Dead/l4dserver
@@ -4,7 +4,7 @@
 # Author: Daniel Gibbs
 # Contributor: Summit Singh Thakur
 # Website: http://gameservermanagers.com
-# Version: 090515
+# Version: 260515
 
 #### Variables ####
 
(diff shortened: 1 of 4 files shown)


dgibbs64:
Wow great work. Thanks for these fixed that's awesome :D. I'm unsure if requiring bash v4 would be an issue, I will check this out before merging

dgibbs64:
just checked this. looks like bash v4 has been a standard for a while. Shouldn't be an issue :D

dgibbs64:
Happy with code merging. Good work :D

xFaNaTiix:
Thank you  =)

Issue #358, opened 2015-05-03
[terraria] Server Request


caycehouse:
I think it would be a great idea to add Terraria to the list of servers, a lot of people like Terraria and it would be something great to add.

xFaNaTiix:
I guess, you have to use TShock (NyxStudios/TShock) for linux. So you have to use mono. I think we should wait for the offical linux support in the new version ( http://terraria.gamepedia.com/1.3 ).

caycehouse:
Yes, I remember using Tshock a while back I forgot about the mono dependency, it probably would be a good idea to wait for native support, thanks for the feedback.

dgibbs64:
Will close for now. However if Native linux support is added I would be happy to give it a go :)

xFaNaTiix:
Just a small update:
"For the long-suffering Mac/Linux users: the plan is to release 1.3 for those platforms by the end of July. The slight delay is to ensure that there are no major issues with the general release before pushing it out to brand new platforms. It's still coming, we promise :)"

Xylemon:
Can this be reopened? There is a Linux server with the latest release. I believe it is able to be downloaded through SteamCMD but the developers are looking into releasing it outside of Steam.

jhg21:
It's no longer in beta, by the way!
https://forums.terraria.org/index.php?threads/terraria-1-3-0-8-can-mac-linux-come-out-play.30287/

dgibbs64:
This will be released with the development branch

dgibbs64:
Holding until official released on the website

dgibbs64:
Finally got around to finishing this off. Enjoy! And please report any issues https://gameservermanagers.com/lgsm/terrariaserver/

PR #478, opened 2015-09-07
Terraria Support

SingingTree:
Add Terraria support as per #358. Tested on OpenSuse 13.2.

Note: A copy of Terraria is required to run the server.


dgibbs64:
Thanks for this. A great help !!!!!

PR #494, opened 2015-10-13
Terraria config install

SingingTree:
This should make the Terraria server move its default config into place correctly. The default config doesn't do anything, but provides a template for others to easily change.

I've tested this as best I can, but since the scripts rely on the config being in the linuxgsm repo, this can only be fully tested once the PR is merged.


SingingTree:
Realised that there was some missing config for the details for Terraria, piggy backing that onto this PR as part of Terraria related cleanup.

dgibbs64:
Merged to master with https://github.com/dgibbs64/linuxgsm/pull/504

PR #511, opened 2015-10-30
gmod-content-mounting

dgibbs64:
Merging this to work in to master


peterjeschke:
Did this PR end up in master? Is there anything blocking it?

UltimateByte:
Nope. For now, I personally believe it's better to make an SRCDS user, then add content into it and use it for all of your gmod servers :)

PR #525, opened 2015-11-03
External per-instance configuration files

jaredballou:
I created a version of insserver to support putting configs that allow overrides of script settings for each instance in a way that makes sense (at least to me). This allows users to simply update the script from GitHub or via download, and not have to back up or re-configure their settings. It also supports multiple instances a little better. I had all this in version 200914, I ported it over to the latest version and tested it with my servers. As far as I can see, it works, but this is my first time doing a PR so be aware that I am pretty new at this.
- Added external config support, load files from cfg/servers directory, in this order:
  - _default.cfg (which is managed by Git and should not be edited by end users)
  - _common.cfg (which is the user's common settings for all servers)
  - SERVICENAME.cfg for the individual instances.
- TODO: If this PR is accepted, modify the default config creation to use a fetch from GitHub as the functions files are.
- Cleaned up the function script calling, it no longer has to change directory when function files are downloaded, which annoyed me.


dgibbs64:
Im liking these changes. For me to merge I will have to complete the changes to all servers. I have however been thinking about this type of feature for a little while and how to better improve updating, loading configs and multiple instances. I wont be merging this however this code will be very useful to me when I start making the major changes and will probably be forking this in to a new branch when i start development.

Thank you for your efforts :D

Insurgency/insserver
```diff
@@ -3,46 +3,87 @@
...
+version="031115"
 
 #### Variables ####
 
-# Notification Email
-# (on|off)
-emailnotification="off"
-email="email@example.com"
+# The name of this script file, used to show the LGSM link properly
+selfname=$(basename $(readlink -f "${BASH_SOURCE[0]}"))
```
dgibbs64: This is a new change. Do you have info on why you choose this?

Insurgency/insserver
```diff
@@ -71,19 +114,19 @@ consolelogdate="${consolelogdir}/${servicename}-console-$(date '+%d-%m-%Y-%H-%M-
 # Do not edit
 
 fn_runfunction(){
+functiondir="${rootdir}/functions"
```
dgibbs64: I would like to merge your changes to fn_runfunction. Could you create a branch just for this change and I will merger this. :)

Insurgency/insserver
```diff
@@ -71,19 +114,19 @@ consolelogdate="${consolelogdir}/${servicename}-console-$(date '+%d-%m-%Y-%H-%M-
...
+functionpath="${functiondir}/${functionfile}"
 # Functions are downloaded and run with this function
-if [ ! -f "${rootdir}/functions/${functionfile}" ]; then
-	cd "${rootdir}"
-	if [ ! -d "functions" ]; then
-		mkdir functions
+if [ ! -f "${functionpath}" ]; then
+#	cd "${rootdir}"
```
dgibbs64: please remove this if not required when creating a PR

Insurgency/insserver
```diff
@@ -3,46 +3,87 @@
...
+lgsm_version="031115"
+logdays="7"
+mapcyclefile="mapcycle.txt"
 maxplayers="16"
+playlist="custom"
 port="27015"
 sourcetvport="27020"
-clientport="27005"
-ip="0.0.0.0"
+srcds_parms=""
```
dgibbs64: Interesting idea adding srcds_parms to separate generic srcds parms and standard parms.

jaredballou:
Glad it could be useful, let me know if you need a hand with anything. Might I also recommend using some sort of versioning header at the top of your functions files, and then have a "script-update-check" and "script-update" actions that compare those versions with the latest ones on GitHub, and download the new versions as required? Or do you already have some method of keeping those files in sync?

Insurgency/insserver
jaredballou: Because this will find the real file name of the script, even if a user creates hardlinks instead of symlinks for the multiple instances.

Insurgency/insserver
jaredballou: I will do this later tonight. I actually have a slightly more involved version that abstracts the downloader a bit more, allowing fetching of not only functions but any files, and I have the rudimentary versioning system for the function files. I'll make one branch for just the functiondir fix to keep it small, and then later you can take a look at the other downloader I have and see if it looks good to you.

Insurgency/insserver
jaredballou: You mean the commented line, or the way I made the function pwd-agnostic?

dgibbs64: just the comment :)

Insurgency/insserver
dgibbs64: Happy with that!

Insurgency/insserver
dgibbs64: Cool will have to get this altered in all the scripts

dgibbs64:
I dont have a function checking tool to update the functions as the functions are quite small. Currently i have an option ./csgoserver update-functions that simply deletes all the functions and it will re-download them. Not very sophisticated but works well haha.

Insurgency/insserver
jaredballou: Thanks, I had that as an addon, but I just realized I should have put the reference to that variable in fn_parms after the last -parm and before the first +parm so that it executes in proper order. Or does srcds just know how to sort that out itself?

Insurgency/insserver
jaredballou: Haven't had a chance to create a branch for just this, but take a look at the way I did it.
https://raw.githubusercontent.com/jaredballou/linuxgsm/master/Insurgency/insserver
Please let me know if you want me to include the version checking (or if you want to do it another way), and I'll try to get a version put up this weekend with just the change for the downloader.

dgibbs64:
Playing with your changes and I like how it works. Just got to figure out how to integrate this for all servers. I like how you have done configs for instances. I think it will work really well. I think the next thing that I need to look at is being able to do something like ./insserver start ins-server2 to save duplicating the main script file and allow a user to select which instance to control.

jaredballou:
I think the symlink method is a better way to handle multiple instances, honestly. Otherwise, you'll need to parse command line options, which opens up a whole new set of problems to solve and input validation issues.

I also thought it might be a good idea to do a pretty major refactor around the scripts themselves, so for at least srcds-based games there would be one "gameserver" script, and then there would be a cfg/games directory which contained all available games. If the script detected it was named "insserver" then it downloads the Insurgency game settings, and goes from there. I'm not good at explaining it, but basically I was looking at removing all the duplication of having each game have a complete instance of the script (even if it's just the frame into which functions get loaded) and make adding new games a little easier.

dgibbs64:
I have been thinking about creating one lgsm-server script. I have really been struggling on how to implement this. I would like to talk to you about this further. I have a lot on currently but at some point I think a chat would be a great idea :)

jaredballou:
Sure, I can help with that. Let me know when you have time, or I can create a Wiki article with my ideas and we can collaborate there? My luck has it when you get free time I'll be slammed and unable to help much, so if it works for you just scoping the problem and getting some basic specs drafted will let me experiment a bit.

dgibbs64:
Sounds great to me :+1:

dgibbs64:
I have merged a modified version of your github branch select. I had to take out the update checker as the function files have different versions to the main file. I will also be using your ideas for managing multiple instances. I am looking at moving from bash to Python. So a new feature that I would like to implement on release of "lgsmPy" this would have multiple instance support. Rather that have to add this feature in bash then re write it in Python I would rather write it in python only.

dgibbs64:
Created a diagram on how I think it should work

http://i.imgur.com/ivm5gJa.png

The problem to solve now is how these multiple instances will be controlled.

It would be easy to control all at once it would be a simple as ./csgoserver start however if the user wanted to control a specific instance. For example if one instance was playing up and needed restarting without killing the other instances. Im not sure how that could best be dealt with.

jaredballou:
So my thinking is, the symlink method is -still- the best way to handle this. Unless you're thinking a person might have one centralized deployment of the insserver script, say in /usr/local/bin, and then want to call it from there, I'm not sure why the symlink method isn't the way to go. It allows a very simple method that anyone off the street can understand, and the alternatives all introduce complexity that I personally think is counterintuitive.

Suppose we make a single script, and then pass the name of the server instance as a parameter, all we're doing is foregoing the symlink in exchange for making every single command more complex. I also think a nice side effect of the symlink method is that adding a new instance is trivial, and is fairly error proof. The alternative is we'd have to build a "create new instance" feature into the manager script to build the blank configs and do all the checks, which users might like but it again increases the complexity of the script. I think there are some definite places to make improvements here, but I don't see how the multiple instances can really be improved in a way that warrants the rework.

We'd also have to either have a backwards compatibility/deprecation function for the legacy installed base, or simply abandon those users by moving to the new system. Having been involved in software development for 15 years now, I've learned to not break UX unless there is a darn good reason backed up by a technical or procedural need, and I don't see that being met here. It's your project, and I'm not so opposed that I'd break off a separate fork or stop helping here, but I just think we may want to be a bit more cautious in how we decide changes for the roadmap.

As fot the Python rewrite, I'm of two minds here. On one hand I have been getting more into Python and its data structures and conditional function declaration in particular make this really appealing to me on a technical level, given how many "hacky" things I've had to do in Bash in the past. On the other hand, that now increases the bar of requirement for users to deploy it, Python is pretty universal but any time we make a change so central to the project we need to think a bit more carefully. Does the increase in fuctionality and maintenance warrant a complete rewrite? Maybe, but then we could rewrite it in Bash in a more correct manner as well, so we need to decide if a Python port in and of itself gives us an advantage.

Perhaps it might be a good idea to have an opt-in "environment report" as part of the current script, and collect and report back anonymized statistics such as Python version or other potential dependencies we'd need? I'd assume you'd port the config format to INI/YAML/JSON, so we'd need to account for the libraries needed to support those files. A complete roadmap of the planned new features would be a good place to start, and then we can work backwards to see what the complete dependency tree would look like, and make sure that doesn't preclude a significant portion of the users. If it does, we ought to at least consider shipping a tool with the LGSM to help get those dependencies met, and assist porting legacy deployment to the new model. Which again increases complexity, and starts the cycle anew. I'm not sure what to do here, and my default mode is to not make a change unless I am certain it is needed.

My advice here would be to engage the users in any way you can and get them to buy in, I've been working on many open source projects that have failed because we as the developers assumed what we were working on was right, when it really might have been what was fun or technically interesting. We'd gloss over issues with user experience to work on something that we assumed was important, and it ended up biting us in the collective ass. So we might be talking about fixing problems that seem to us as the authors as a big deal because we see the problems and deal with them on development, but to make a change that increases the USER'S difficulty in using the tool to solve our own difficulties is a dangerous path to go down. That doesn't seem to be the case here, but having been in the industry this long my mind functions in worst case scenarios and paranoia as a default, so as long as you consider these issues it should be fine.

The end user experience must be the primary concern of the architects, in looking at your diagram it is appealing to me on a technical level, but from a user perspective I believe it'll confuse people, particularly the legacy users that we have now. From my perspective as a potential contributor and very happy user of your product, I'd ask what problem making this major change solves for me? If there is a good answer to that question, then we should pursue it and map out the actual end-product value of these changes, so we can decide if it's worth all the effort to make this major revamp. If it's not, we should focus on making the existing tool more robust and easy to use.

Please take this all as feedback from a person who is personally invested in your project, and wants more than anything to help make it a success. If you'd like to move this discussion to a different venue and get more input from others, I think that might be the best next move here. At the least, I personally would like to get more feedback from other users to see what sort of issues or suggestions they may have, and refocus my own efforts accordingly.

dgibbs64:
@jaredballou thanks again for the in depth feedback I can tell you really care about the project. Again I will need to sit down have have a good read. Im sure I am going to learn a lot :+1:

dgibbs64:
@jaredballou My diagram is intended to give me an idea on how to layout LGSM for multiple instances using your idea....at least how I perceive it (because I am struggling to figure out simple implementation UX wise). On the right side is more what you suggest with managing multiple instances.

on the left is something some users have requested. I see it as a deployment script (initially) or even an overall controller program for managing multiple games (maybe). That will allow deployment of the various game servers by using a command or a menu. Similar to the below script I use for deploying demo servers for development (rsync serverfiles to VPS).

capture

If i was to do something like that then I wouldn't want it to change how 'legacy' users deploy servers i.e one server at a time. But rather allow power users that run multiple games to simply add a server automatically by selecting it from a menu.

I think what I am trying to achieve is to retain the current simplicity of LGSM but allow power users more options (which can add complexity) for deployment and management on a bigger scale. This is why I have struggled with multiple instances for a long time as it adds a level of complexity I have struggled to simplify.

My mindset is definitely 'Keep it simple stupid' when it comes to this project. Which I believe has made it so popular.

jaredballou:
I think this is a good example of a product split being wise. So, we have the core "run the server" script, now people want to orchestrate it? Excellent, so let's define the features that are wanted in the orchestration tool, and create a spec for that. Then, as new features come in, we decide if it's orchestration or for servers themselves, and then put the work in the right column. It might sound like more work to split it up this way, but not only is it the "UNIX Way (tm)" to split programs into smaller sub-components, you'll find that it makes maintenance a lot easier. When we can encapsulate changes, it makes the testing and debugging MUCH easier, especially once both projects become fairly large and complex. The more you can do to "gate off" change and be able to control the pace of modification, the better you can stand behind it. I'd be more than happy to help you design a plan around this, my experience is more in the implementation side but I've assisted on enough design meetings for successful products that I'm sure we can do something to set the stage for success here.

dgibbs64:
I finally understand your method and its so simple!!!! Im developing this currently

dgibbs64:
thanks @jaredballou this branch was the inspiration for the new improvements to LinuxGSM

PR #553, opened 2015-12-12
Multiple server instances code

dgibbs64:


dgibbs64:
A version of LinuxGSM has now been released that was inspired by @jaredballou original idea.

PR #569, opened 2015-12-21
enhanced mod-support

PhilPhonic:
added parameters. descriptions taken from: https://community.bistudio.com/wiki/Arma_3_Startup_Parameters

-servermod
Loads the specified sub-folders for different server-side (not broadcasted to clients) mods. Separated by semi-colons. Absolute path and multiple stacked folders are possible. 

-autoinit:
Automatically initialize mission just like first client does. Note: Server config file (server.cfg) must contain Persistent=1; , if it's 0 autoInit skips. 

-loadmissiontomemory: 
Server will load mission into memory on first client downloading it. Then it keeps it pre-processed pre-cached in memory for next clients, saving some server CPU cycles

### DIFF
--- a/Arma3/arma3server
+++ b/Arma3/arma3server
@@ -25,10 +25,11 @@ steampass="password"
 
 # Start Variables
 ip="0.0.0.0"
+srvport="2302"
 updateonstart="off"
 
 fn_parms(){
-parms="-netlog -ip=${ip} -cfg=${networkcfgfullpath} -config=${servercfgfullpath} -mod=${mods} -servermod=${servermods} -bepath=${bepath} -autoinit -loadmissiontomemory"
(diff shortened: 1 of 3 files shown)


dgibbs64:
Looks good to me. I have little knowledge of ARMA 3 so I am relying on your expertise on this :).
The only thing that may need looking at is ./arma3server details the port may not show correctly now as you have moved it to the start parms. 

If you are recommending that ports are dealt with in the start parms im happy to use this as default. But I would comment out this on the default config and leave a note reccomending using start parms over a config

PhilPhonic:
You are right..
**./arma3server details** gives me:
> DESCRIPTION              DIRECTION  PORT  PROTOCOL
> Game                   INBOUND    0     udp
> Steam: Query           INBOUND    0     udp
> Steam: Master traffic  INBOUND    0     udp

The problem is, it only seems to work when added as a parameter. When added to config, it gives me this weird ports seen in my last post.

dgibbs64:
Happy to merge this. I will need to review it and then merge to bare with me as may take a few days :)

PR #578, opened 2015-12-25
support for MariaDB/MySQL

PhilPhonic:
I added support for installing TS3 with MariaDB/MySQL instead of SQLite.


dgibbs64:
Decent! Does the user have the option to select which dB they prefer mysql,mariadb or sqlite?

PhilPhonic:
yes. during installation process, the user is able to choose between sqlite or mariadb plugin. the mariadb plugin is also compatible to mysql because mariadb is a fork of mysql

PhilPhonic:
important to mention: libmariadb2 has to be installed to use the mariadb plugin!
on debian: **sudo apt-get install libmariadb2**

Issue #418, opened 2015-06-12
[DST] Server Request


marvinlehmann:
Don't Starve Together: http://store.steampowered.com/app/322330/
Useful links:
http://forums.kleientertainment.com/forum/73-dont-starve-together-beta/
http://dont-starve-game.wikia.com/wiki/Guides/Don%E2%80%99t_Starve_Together_Dedicated_Servers

dgibbs64:
I can certainly give it a go I would need to read up on the server etc and may take some time as its not an engine I have done before. I also don't own the game so would also need this game donated.

xFaNaTiix:
Donated! =)

dgibbs64:
Thanks :D. I look forward to putting this together

dgibbs64:
./dontstarve_dedicated_server_nullrenderer: error while loading shared libraries: libcurl-gnutls.so.4: cannot open shared object file: No such file or directory

Currently getting this issue while trying to start the server on both Debian and CentOS

BlackDwarfUK:
As per http://dont-starve-game.wikia.com/wiki/Don%E2%80%99t_Starve_Together_Dedicated_Servers, libcurl4-gnutls-dev:i386 is a pre-req for the server on Debian.

dgibbs64:
The server is released but not working in centos and maybe Debian...I cannot remember. either way I cannot resolve the issue until the game developers come up with something. I will keep an eye on this.

mlopezv86:
I finally managed to install tmux on CentOS 6.2 and got the following error:

[ FAIL ] Starting dst-server: Unable to start dstserver: Tmux error:

Command
=================================
tmux new-session -d -s "dst-server" "./dontstarve_dedicated_server_nullrenderer "

Error
=================================
tmux: error while loading shared libraries: libevent-2.0.so.5: cannot open shared object file: No such file or directory

Fix
=================================
No known fix currently. Please log an issue.

Then, fixed it by creating the appropriate symlink: ln -s /usr/local/lib/libevent-2.0.so.5 /usr/lib64/libevent-2.0.so.5 (ln -s /usr/local/lib/libevent-2.0.so.5 /usr/lib/libevent-2.0.so.5 for 32-bit systems).

Tried to start again the server and got the following:
[ .... ] Starting dst-server: dstserverno server running on /tmp/tmux-500/default
[ FAIL ] Starting dst-server: Unable to start dstserver

[...]

PR #601, opened 2016-01-02
Don't Starve Together CentOS Fix

dgibbs64:


dgibbs64:
CentOS now working with a fix #418

PR #609, opened 2016-01-06
insserver More config options

PhilPhonic:


dgibbs64:
I think you might like this repo. Could you do a pull request for your config file changes on this repo as well? https://github.com/dgibbs64/gameserver-configs

PR #614, opened 2016-01-07
Ts3 password

UltimateByte:
It should be fine :)


dgibbs64:
I will test and merge

Issue #629, opened 2016-01-14
[insserver] Part of fix_ins.sh may no longer be needed


dgibbs64:
# fix for issue #529 - gamemode not passed to debug or start

	if [ "${function_selfname}" == "command_debug.sh" ]; then
	        defaultmap="\"${defaultmap}\""
	else
	        defaultmap="\\\"${defaultmap}\\\""
	fi
	fn_parms

I am using insserver currently and found that using this fix appears to now break debug mode. Can anyone confirm this and I will remove this fix. Maybe an update to Insurgency resolved the issue.

PhilPhonic:
I can confirm this.
:~$ ./insserver debug
> map load failed: \ not found or invalid

If you remove the fix, gamemode won't be passed to ./insserver start or ./insserver debug
./insserver start still works fine with this fix.

I'll try to figure out what's wrong

PR #634, opened 2016-01-17
[insserver] fix start/debug

PhilPhonic:
fix for issue #629


PhilPhonic:
I have no idea why calling "ins_params" was needed, and now no longer is..
With this fix both start and debug work fine for me.

PR #649, opened 2016-01-25
adding Sven Coop 5.0

hurdad:
Ubuntu 14.04 LTS 32-bit


dgibbs64:
I love this type of pull request :-). I'm going to need to implement a better file download function to handle non steam files. I will look at this later

hurdad:
they just added sven cop to steam [http://store.steampowered.com/app/225840/]

dgibbs64:
O well thats OK then ;-P

olliee:
I see the SvenCoop folder, but not documentation and I tried running myself however it is having an error with permissions check. Has this been implemented or is it still in progress?

hurdad:
what error are you getting? Should be working. I think there is an additional dependency for sven coop : libssl1.0.0:i386

olliee:
@hurdad http://prntscr.com/blfwx0

hurdad:
looks like you have an OS with an older version of GLIBC. What OS are you running? Sven Coop requires 2.18+

https://github.com/dgibbs64/linuxgsm/wiki/Glibc#server-requirements

Your going to need to use Ubuntu 14.04+, which has GLIBC (2.19)

PR #668, opened 2016-02-01
monitor start

PhilPhonic:
ability to start a server if there is no lockfile

### DIFF
--- a/7DaysToDie/sdtdserver
+++ b/7DaysToDie/sdtdserver
@@ -9,7 +9,7 @@ if [ -f ".dev-debug" ]; then
 	set -x
 fi
 
-version="271215"
+version="010216"
 
 #### Variables ####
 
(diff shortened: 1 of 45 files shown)


PhilPhonic:
Couldn't test all servers, because I don't own all the games.
I think this should work for every game..

dgibbs64:
Looks like this is a lockfiles bypass. What is your main reason for wanting this? The lockfile prevents monitor from starting back up if the user specifically stopped the server.

PhilPhonic:
The lockfile still can prevent monitor from starting the server when setting startonmonitor="off" (default).
There is no "reason" why I'm wanting this feature, I just thought it would be helpful. Perhaps there is a constellation of thingst that can happen, where the server crashes and the lockfile is removed? In this case monitor would restart the server.
If there is no way this could happen, just deny my pull request ;)

dgibbs64:
the lock file is only removed on stop and created with start. The only way there could be an issue is if the user deleted the lock file and the monitor would stop working.

If the server crashes and reboots the lockfile still exists so monitor will then start the server again.

I will deny this pull request on this occasion :)

PR #688, opened 2016-02-10
Logfiles cleaning

### LINKED ISSUE #354: Server logfile auto-delete (several GB after a few months)
Hi there,

I'm still using this amazing script that is wonderfully improved every time i need to download it again, and i'm very thankful for that !

By the way, i just migrated a few servers these days, and my gosh, i noticed that server logfiles aren't auto removed ! On our biggest servers on Gmod, it was more than 4GB of logiles in /log/server and a lot of files, so that was the longest thing to copy from one server to the other !
I was just thinking : why not auto-remove those just like you're doing with both the script and console logfiles ?

Also, as a bonus, it could be nice to be able to setup the amount of logfiles we want to keep into the "gameserver" script.

Thanks for your attention.

PS : Kept a screenshot of the sledbuild server's logfiles... :p http://image.noelshack.com/fichiers/2015/18/1430193284-logiles-size.png

UltimateByte:
fixes #354 

Took me some time to figure out how it was working, and to simplify it a little bit (with the exact same functionalities), but i think it's way cleaner that way.
Tested, working on one of my DarkRP.
I see no reason why it wouldn't work for sourcemod, as the folder path is right, and i triple checked the syntax.

### DIFF
--- a/functions/logs.sh
+++ b/functions/logs.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 # LGSM logs.sh function
 # Author: Daniel Gibbs
+# Contributor: UltimateByte
 # Website: http://gameservermanagers.com
 lgsm_version="100215"
 
--- a/functions/logs.sh
+++ b/functions/logs.sh
@@ -15,17 +15,10 @@ if [ -n "${consolelog}" ]; then
 	fi
 fi
 
-# Set source logs directories
-if [ -z "${systemdir}" ]&&[ "${engine}" == "source" ]; then
-	srcdslogdir="${systemdir}/logs"
-	# Set addons directories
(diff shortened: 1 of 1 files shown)


UltimateByte:
Confirm working with CSS, it took a while, lol... I never cleaned it before.

[  OK  ] Log Manager css-server: Starting
[ INFO ] Log Manager css-server: Removing logs older than 7 days
[  OK  ] Log Manager css-server: Removed 41382 log files

I also saw a /log/server folder, as my installation was pretty old. I'll make a function to clean those too.
Hopefully i can change the code after pull requesting.

UltimateByte:
Here it is, the old /log/server folder will be cleaned, and removed only if empty (means containling no log since ${logdays}.

PR #706, opened 2016-02-18
Accept two main other mail servers

UltimateByte:
In the end, i went for checking that a mail server is here in any case.
Before the last commit, it wouldn't check packages as long as a mailx folder was detected, which wasn't very clean.

### DIFF
--- a/functions/check_deps.sh
+++ b/functions/check_deps.sh
@@ -40,10 +40,16 @@ fi
 fn_deps_email(){
 # Adds postfix to required dependencies if email notification is enabled
 if [ "${emailnotification}" == "on" ]; then
-	if [ -n "$(command -v dpkg-query)" ]; then
-		array_deps_required+=( mailutils postfix )
-	elif [ -n "$(command -v yum)" ]; then
-		array_deps_required+=( mailx postfix )
+	if [ -d /etc/exim4 ]; then
+		array_deps_required+=( exim4 )
+	elif [ -d /etc/sendmail ]; then
+		array_deps_required+=( sendmail )
+	elif [ ! -f /usr/bin/mailx ]; then
+		if [ -n "$(command -v dpkg-query)" ]; then
+			array_deps_required+=( mailutils postfix )
+		elif [ -n "$(command -v yum)" ]; then
+			array_deps_required+=( mailx postfix )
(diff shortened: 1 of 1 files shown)


UltimateByte:
Also, maybe would be cool to add the mail package check on start, and just put a warning, if mail server isn't detected, as people often set this after the installation.

dgibbs64:
Do these mail servers use the mail command? As that is what LGSM uses

UltimateByte:
I found : 

Syntax for exim : `mail email@domain.com -s mail-title`
Syntax for sendmail : `# echo “This will go into the body of the mail.” | mail -s “Hello world” you@youremailid.com`

Seems like they're pretty similar, it's just a matter of how those examples were written.

Seems like all of them require mailx anyway, that must be the one providing the common `mail `command.

UltimateByte:
Also found out here http://www.linuxfromscratch.org/blfs/view/cvs/basicnet/mailx.html

That 

> mail is a symbolic link to mailx.

dgibbs64:
Im guessing users have been trying to use non postfix mailservers and thats why you are working on this. This is a pain as Its another thing to try and support. I know very little about sendmail and exim4.

If we are to add support for non postfix mail server then Maybe the code should instead look for existing mail servers in check_deps and only reccomend postfix if missing or already installed. Also the email feature would also need to detect sendmail exim4 and postfix then using the relevant command to send an email.

UltimateByte:
You're guessing right 
https://github.com/dgibbs64/linuxgsm/issues/695

Currently, what it does is : If sendmail is installed, it requires sendmail, if exim4 is installed, it requires exim4, otherwise, the behavior isn't changed. Does the mail command changes depending on the mail server ? Seems like they're compatible together.

dgibbs64:
as per @jaredballou  checking for /usr/bin/mailx will probably be the best idea. Should this fail suggest installing postfix. That way if users have an alternative server and mailx is installed that shoudl be OK too.

UltimateByte:
OK, i'll change that

UltimateByte:
Here, should be fine. Tested, successfully had the right required dependencies onto my server.

Would be great to have it running while using email-test though.
I'll try that.

PR #709, opened 2016-02-20
FastDL first release

UltimateByte:
Working for gmod only right now, will add some other source games soon


UltimateByte:
Not sure why the last test failed, makes no sense :o))

PR #718, opened 2016-02-26
Update on start variable add

WSG-ViViV:


UltimateByte:
You're damn right !
After experimenting with my own server, i'd add server.seed to fn_parms too, but i'm not sure what it does if left blank. What do you think ?

PR #724, opened 2016-02-28
Add seed map

ivangc11:
Add the possibility that the user enter the seed that wants to generate map. Currently it is being generated 691977496 seed, and always the same seed is generated.

### DIFF
--- a/Rust/rustserver
+++ b/Rust/rustserver
@@ -34,13 +34,14 @@ rconpassword="CHANGE_ME"
 maxplayers="50"
 
 # Advanced
+seed="691977496" # Map seed
 worldsize="4000" # default 4000; min : 2000 max : 8000 
 saveinterval="300" # Auto-save in seconds
 tickrate="30" # default 30; min acceptable 15 , max 100 or 128
 
 # https://developer.valvesoftware.com/wiki/Rust_Dedicated_Server
 fn_parms(){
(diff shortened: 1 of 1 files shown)


UltimateByte:
The proper command is +server.seed

I didn't put it yet, because the default value should be blank, and i'm not sure about the behavior when the command is present but blank. I need to test it out first.

UltimateByte:
Testing it out right now

UltimateByte:
So

* without +server.seed , i had a map with a seed of Map_4000_493175682, so no, it's not always 691977496, it's just random between acceptable values.

* With a blank +server.seed , i had a map with a seed of 1.

=> So a blank +server.seed changes the randomness of the map to a default of 1, which isn't good.

What we can do to counter this, is add a function checking if the seed variable is blank, and add the `+server.seed ${seed}` start parameter only if it's not blank.

dgibbs64:
You want this merged?

ivangc11:
if not add variable `+server.seed ${seed}`, I always generates the same seed, in my case 691977496.
Can you check if you you always generates the same?
I stop the server, delete the save folder containing the map, and start it always generates the same map.

UltimateByte:
@ivangc11 Nope, you say that because you didn't remove your previous map before. If you already have a map in your server identity folder, it keeps it, whatever the server.seed is. The behavior i stated before is accurate, i've tested it this morning, when nobody was onto my server..

@dgibbs64 
Merge my pull request, ;)
https://github.com/dgibbs64/linuxgsm/pull/725

UltimateByte:
You may be right for the random value being saved. But a blank "+server.seed" generates a value of 1. And this is bad. That's why this start option must be conditional : add it only of "seed=" isn't blank. Already pulled. Thanks for giving me the motivation to test it out asap btw. :)

UltimateByte:
No way we merge it with a default seed number btw, but thanks again for suggesting, it's very relevant to get it added. Closing :)

UltimateByte:
See #725 to be merge soon

PR #725, opened 2016-02-28
conditional server.seed and comments rework

UltimateByte:
Can't test it out as multiple server instances don't work and my server now has players on it, but if that looks OK to you, it should be fine.

### DIFF
--- a/Rust/rustserver
+++ b/Rust/rustserver
@@ -45,7 +45,7 @@ parms="-batchmode +server.ip ${ip} +server.port ${port} +server.tickrate ${tickr
 }
 
 # Specific to Rust
-if [ -n ${seed} ]; then
+if [ -n "${seed}" ]; then
 	# If set, then add to start parms
 	conditionalseed="+server.seed ${seed}"
 else
--- a/Rust/rustserver
+++ b/Rust/rustserver
@@ -34,14 +34,24 @@ rconpassword="CHANGE_ME"
 maxplayers="50"
 
 # Advanced
-worldsize="4000" # default 4000; min : 2000 max : 8000 
+seed="" #  default random; range : -2147483647 to 2147483647 ; used to change or reproduce a procedural map
+worldsize="4000" # default 4000; range : 2000 to 8000 ; map size in meters
 saveinterval="300" # Auto-save in seconds
-tickrate="30" # default 30; min acceptable 15 , max 100 or 128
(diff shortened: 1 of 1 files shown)


ivangc11:
You can check you out random seed if you do not enter the variable? My always generates me the same.

UltimateByte:
As i told you : No seed = random value
Empty server.seed = 1

Of course if you start your server once, the random value will be keeped until you remove your files in your server identity folder (in serverfiles/server/your_identity). For your information, the identity is changed by changing "servicename" variable.

UltimateByte:
Tested working, can merge

ivangc11:
I can not get the seed out randomly. Stop the server, delete the save folder in the path of my server, in my case is: /serverfiles/server/rust-server/save. I've also tried to remove Storage.db and UserPersistence.db files.
When I return to start the server, I generated the map with the same seed.
What am I doing wrong?

Another question, the file you have modified the github, I have to download it manually or automatically download in any way ?.
a greeting

UltimateByte:
This file has to be downloaded manually, as it's the main script. It's the only one that needs to be downloaded manually, otherwise, you can run the update-functions command. https://github.com/dgibbs64/linuxgsm/wiki/Update#updating-your-lgsm-script

It's possible that this random number is random to your machine. I'm not so sure about that. But the purpose of this little add isn't wether or not it's random, it's if you specify +server.seed with no value, it will always take the value of 1, and that is not desirable.

ivangc11:
But the steps I performed to erase the data, correct? How do you do it?
Thank you

UltimateByte:
Save your rustserver settings in a local file
```
rm rustserver
wget http://gameservermanagers.com/dl/rustserver
nano rustserver
```
Edit rustserver to fit your needs.

ivangc11:
thanks friend, but I mean to delete data from the saved game. What do wipe typical when updates.

UltimateByte:
The forced wipe happens tomorrow, i'll have to figure this out at the moment. 
I did it once, but i'm not so sure if i was right : 

/serverfiles/server/rust-server
Remove UserPersistence.db / Storage.db / save folder
One of them contains blueprints from players and doesn't need to be wiped all the time.

ivangc11:
Ok, I do the same. I have added to steam, aceptame and we tried talking and servers. a greeting.

PR #754, opened 2016-03-13
Gmod tickrate and fix for many maps

UltimateByte:
Fixes https://github.com/dgibbs64/linuxgsm/issues/685
Tickrate needs to be lowered in order to be able to handle a large amount of players.
Default tickrate is 66.66, but the -tickrate commnand will round numbers to the lower integer. So i'll default at 66 and not -the more accurate- 67 to not worry anyone with a "non standard" value.

For the +r_hunkalloclightmaps 0, it fixes the hunk overflow crash on server startup when using maps with too much lights (and there are a lot of them). It changes the way lights are stored in RAM, the 0 mode (default one a few years ago, and still default one on clients) can handle a way larger amount of lights, but a theoretically a bit slower way. I noticed no difference, so it's a good thing to fix it by default.

### DIFF
--- a/GarrysMod/gmodserver
+++ b/GarrysMod/gmodserver
@@ -9,7 +9,7 @@ if [ -f ".dev-debug" ]; then
 	set -x
 fi
 
-version="190216"
+version="130316"
 
 #### Variables ####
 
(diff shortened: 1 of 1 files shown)


UltimateByte:
Can't pull request to 2 branches at the same time, but would be cool to pull to the development branche afterward.

PR #765, opened 2016-03-18
Fix more TS3 change password related issues

drantor:
Always start the server with inifile parameter to ensure right settings.
And reset the ts3serverpass variable so it won't restart the server again in new password mode.


dgibbs64:
Thanks :)

PR #770, opened 2016-03-24
Add common parameters to Ark start

yadutaf:
Hi,

This PR fixes a Typo in the contributor URL and adds a couple of parameters to the Ark launcher based on the params of Rust. The most important was the ``ip`` as the launcher refuses to start if it is not set but did not take it into account. ``rconpassword`` is empty by default which seems to effectively disable it.


dgibbs64:
Currently the script uses the config file for these parameters. Is there a reason why you shoudl use the parms instead of the config file?

yadutaf:
Regarding "ip" param, if it's not set you get a warning message BUT it is ignored when set.

For all other params, this is mostly convenience. linuxgsm offers a simple / standardized way to configure most common parameters. So, it "felt right" this way. Anyway, I only used the "ip" in my scenario.

dgibbs64:
Thanks. I normally have to choose one or the other. A few servers have the option of either. This makes things complicated. I will have to think of a way to best deal with this situation.

cedarlug:
It seems that the config file does not honor these settings on startup.  If you make the QueryPort something other than 27015 in the GameUserSettings.ini, it has no efffect.

So the startup parameters in the script seems like the right place for them.

This would also pin down Ark's parmslocation as "selfname" in command_details.sh

Etsche:
I have the same . The Scrips dont found the place from the GameUserSettings.ini . 
Change ports by editing the parameters in:
UNKNOWN

dgibbs64:
This has now been merged by @cedarlug thanks for pointing this out the parms method is now the default

PR #789, opened 2016-04-23
Added branch variable

DrCox1911:
Ability to opt in to different branches of a steamgame, e.g. iwillbackupmysave branch from Project Zomboid. To us this you have to add a variable named "branch" to your gameserverscript.


UltimateByte:
Seems cool, but would also require a branch variable to all game script :)

DrCox1911:
Yep, but for the future I think it´s the way to go. Many games use different branches (e.g. Project Zomboid, Arma3, Natural Selection 2, ...).

Maybe some sort of check if the variable is present could be feasible?

UltimateByte:
Just thinking about that, wouldn't it be better to rename the variable to "steambranch" to avoid any confusion with "gitbranch" ?

PR #853, opened 2016-05-31
Add variable to set email's sender 'emailfrom'

jakubsuchybio:
Hi,
on our server the default sender mail was awful, so I tweaked your alert_email, that it can now send emails with given sender set in new property 'emailfrom'
Regards Jacob

### DIFF
--- a/CounterStrikeGlobalOffensive/csgoserver
+++ b/CounterStrikeGlobalOffensive/csgoserver
@@ -19,8 +19,9 @@ version="210516"
 # Email
 emailalert="off"
 email="email@example.com"
+#emailfrom="email@example.com"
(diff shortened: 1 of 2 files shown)


UltimateByte:
I like that, it could prevent spam detection. However, if it works, the variable should be added to all game servers.

jakubsuchybio:
Yes, I thought, that it would need to be applied to all servers. Ok, I will update the PR.
I tried it and it worked. Would you like some proof of it? I don't think I am that good to write some tests, but I can show some pictures of it working. If it would be enough.

dgibbs64:
Is there going to be a pull request for this?

PR #855, opened 2016-06-01
Checks if "log" folder already exists

ScottSmudger:
If the folder does exist, just displays a warning. Otherwise, it creates the folders.

This just removes the standard bash error message from the user.


UltimateByte:
"if [ "${checklogs}" != "1" ]; then"
It's supposed to happen already. The proper solution would be to add a checklogs at the place where you probably had an issue.

dgibbs64:
Thanks for this. However it is not really required for the installer. mkdir will state if the dir already exists. Also your code did not take in to account if only one dir was missing.

PR #861, opened 2016-06-05
Fixed lock file removal for source servers

marvinl97:
fn_stop_graceful_source() didn't call fn_stop_tmux() or ' rm -f "${rootdir}/${lockselfname}" '. That's why monitoring restarted the server when it was shutdown gratefully.


UltimateByte:
Nice found. About that, can we rally call it RCON quit whent it's more of a console command quit ?

PR #881, opened 2016-06-12
Adding in parsing for realvirtual engine - allows arma3server to find…

cedarlug:
… its parameters.

(remove spurious end-of-line whitespace.)


cedarlug:
Breaks ts3server?

PR #882, opened 2016-06-12
allows arma3server to find its parameters.

dgibbs64:


UltimateByte:
Awesome !

igeighty:
this works. 
when performing ./arma3server details
the correct info is shown. in the command line parameters and the ports.
prior to this no info would show in the ports 

http://pastebin.com/MYHSTXMc

thanks for the fix

PR #888, opened 2016-06-15
Update: command_details.sh - Show Pushbullet alert

ki2007:
Show Pushbullet with Email alert under Script Details


dgibbs64:
Thanks!!

PR #889, opened 2016-06-17
enabling automatic dependency installation

hurdad:
tested on CentOS (yum) and Ubuntu (apt-get)


dgibbs64:
Wow I never noticed this error. Thanks for resolving :D

PR #897, opened 2016-06-23
Comment and default map settings cleanup.

lukeacl:
Some of the map settings have different ranges or they've changed in recent updates to Rust. I've cleaned up the settings and adjusted the comments as required.


dgibbs64:
Thanks!

Issue #900, opened 2016-06-25
CSGO Server


geggie88:
hallo i have a problem with my csgo server. 
i installd it with this https://gameservermanagers.com/lgsm/csgoserver/#gettingstarted

so i have the problem that the csgo server no realy lod the mods only one can join you have only one magazin in your gund ( 30/00 ) you cant   see your overview from the game  and and and

dgibbs64:
What are you even talking about?

geggie88:
The csgo server that i install with this programm
https://gameservermanagers.com/lgsm/csgoserver/#gettingstarted
is running but competetiv mod is not working or deathmatch no mode is
working

2016-06-26 9:07 GMT+02:00 Daniel Gibbs <notifications@github.com>:

marvinl97:
It's impossible to help you if don't provide useful information.
What exactly did you do? And what are your config/server settings? Did you already try a debug start (./csgoserver debug) and checked for error messages? And when no mode is working, what the hell is running on the server?! Casual?

Did you provide a Game Server Login Token?
https://github.com/dgibbs64/linuxgsm/blob/master/CounterStrikeGlobalOffensive/csgoserver#L55

UltimateByte:
I'm guessing this is the same issue ?
http://steamcommunity.com/groups/linuxgsm/discussions/0/358415206096429001/
We can't support modding here, if you're using mods, ask for help to this modding community.
If you think there is a problem with LGSM itself, then that's gonna be in the range of what we can help you with.

geggie88:
Okay if you have steam add me please than we can show on my server what i
mean. i dont knoo what mod working on this server for ist that nothing i
set 32 bots deathmatch mode and some other things but when you join the
server you have no knife no weapon you cant see how much money you have und
more issues dont understand thist

hear my steam link: http://steamcommunity.com/id/deadpool1988/

and thank you for helping me

2016-06-27 0:18 GMT+02:00 UltimateByte <notifications@github.com>:

UltimateByte:
Did you set this in csgoserver script ?

```
gamemode="2"
gametype="1"
```

(i won't add you on steam)

marvinl97:
I send him a request. I'm German too. This should make everythink a bit easier.

UltimateByte:
Thanls. Please, keep us tuned on how you solved the issue, it might help us helping other people.

marvinl97:
[...]

PR #904, opened 2016-06-27
Changed the order of game mode variables

marvinl97:
To avoid possible confusion like in #900. The old order encouraged the mix up of "gametype" and "gamemode".


UltimateByte:
That's what i thought too while reading it here https://github.com/dgibbs64/linuxgsm/issues/900#issuecomment-228698016. I made the mistake the first time, but as i double check pretty often i had it right anyways in my post :p

UltimateByte:
I can safely merge this one :)

PR #909, opened 2016-06-30
Fixing the issue when the process id vanishes.

cedarlug:
Cleans up any non-digit returned in the resulting process id query when the process has terminated.

In the event the process id vanishes, the "let" command in the subsequent line sets the value to 0 (which was the intended result when no process id was found).


cedarlug:
ts3server_check fail for the win.

dgibbs64:
lol

PR #942, opened 2016-07-26
Mainly minor fixes

marvinl97:
The only critical one is "Fixed typos" because I changed the gamename "Teamspeak 3" to "TeamSpeak 3". Users would have to update that.

Note to "Added directory check":
Probably not the best solution. Just a really simple one. I don't know if a notification should be displayed to the user?


UltimateByte:
Wow man, did you read the whole code for that amazing work ? :o))

marvinl97:
More or less. I did my best. =)

dgibbs64:
Brilliant work. I see some decent tidying up of the code. I'm very impressed!!

UltimateByte:
I have to say that i don't get this "NOT SET" thing neither.
For the ${unavailable}, it's just a mater of color
`unavailable="${red}UNAVAILABLE${default}"`

I think this function would need a major rework anyways, including an option to hide sensitive information for reporting to forums.

dgibbs64:
NOT SET means that variable is empty. In other words its not present in the config file as the user did not set it.

UNAVAILABLE however means that the config file is missing `[ ! -f "${servercfgfullpath}" ]` and is used as a placeholder. The user cannot set it as there is not config file to add the variable.

Also does that mean that openjdk-7-jre is already installed on Ubuntu 16.04?

Your name change for teamspeak will cause issues for older versions. I think a fix for this would be to add in to core_functions.sh something like

`if $gamename == TeamSpeak 3 then gamename=Teamspeak 3`

to convert the old name to the new one this preventing errors and saving every teamspeak 3 user being forced to update there scripts. This is probably why Travis CI test failed. Please can you try to resolve this and let me know :).

Im looking forward to merging this one. Im very impressed with this merge as it was probably quite tedious work. But I think well worth it!! Thanks again

UltimateByte:
@dgibbs64 Actually, i think it's more about this syntax that we don't get : `servername=${servername:-"NOT SET"}`

@marvinl97 Let''s see if that commit fixes it :)

marvinl97:
I hope so. Watching the travis log/console closely :D

UltimateByte:
`The command "bash tests/tests_ts3server.sh" exited with 1.`

Still caps in here :p 
```
 # Server Details
 -gamename="Teamspeak 3"		 +gamename="TeamSpeak 3"
 -servername="Teamspeak 3 Server"		 +servername="TeamSpeak 3 Server"
```

marvinl97:
I don't get it.

dgibbs64:
I will take a look later on :-)

dgibbs64:
Re running the test. Doesn't make sense why it failed.

dgibbs64:
Failed again :/. Do me a favor and create a new pull request but going to the development branch please. Then I can review this better and merge to master when im happy :)

marvinl97:
Alright. I'll do this tomorrow. This is the best option. =)

PR #947, opened 2016-07-27
Mumble full rework

### LINKED ISSUE #745: [mumbleserver] Stop doesnt work|Missing parameter in script
Hi,
I just installed a new mumble Server and came up with 2 errors.

1. I have multiple IPs on my system so the script ofc asks me to set it to one. But there is no IP line in the script for it. If I insert it by myself it works.

2. If I try to stop the server with the script its says "OK" but the server wont shut down and I have to kill the process manually. 

Additionally When I saw it right, the source for the Linux Static Server files moved to github (https://github.com/mumble-voip/mumble/) instead of Sourceforge
~Flam

UltimateByte:
I've added install, update, monitoring functions, fixed start, stop and details functions...
Everything is now working fine during my testing.

More than fixes #745


UltimateByte:
Note this change to the core_dl, otherwise it couldn't download from github. Took me ages to figure this bad boy out. https://github.com/dgibbs64/linuxgsm/pull/947/commits/b0253a9ff555ad71e326e48a6b73fb079511bbec

dgibbs64:
WOW. I will look at all this work as soon a possible :)

dgibbs64:
I will do a full test and check all the code before a commit. Probably not a good idea to commit when tired. At quick glance looks decent :D

UltimateByte:
You're right :)
Maybe displayed messages could be improved while updating, you'll see.
To simulate an update, like for TS, you can change the version number in the last log.

Saving you some time : 
https://raw.githubusercontent.com/dgibbs64/linuxgsm/mumble-git-dl/Mumble/mumbleserver
Don't forget to change the branch to mumble-git-dl

Crap, speaking of logs, i'm thinking now there won't have any due to the fact that it's not using tmux anymore... Brrrr ! Now i need to find a workaround. I'll probably use tmux anyways then, will have to rework start and stop, again.

lgsm/functions/update_mumble.sh
```diff
@@ -15,7 +15,6 @@ fn_update_mumble_dl(){
 	echo -e "copying to ${filesdir}...\c"
 	fn_script_log "Copying to ${filesdir}"
 	cp -R "${lgsmdir}/tmp/murmur-static_${mumblearch}-${availablebuild}/"* "${filesdir}"
-	rm -R "${lgsmdir}/tmp"
```
UltimateByte: Interesting : Is there a function removing the temp dir that i don't know ? :)

dgibbs64: The tmp dir does not get removed currently. However that may change at some point

dgibbs64:
Im testing this currently. Should be able to merge soon :)

dgibbs64:
Awesome work. Mumble server has never been better!!!

UltimateByte:
Thanks ! It really needed it; most functions were not working, it was pretty much unusable.
My checklist is finally getting smaller and smaller :o))

PR #958, opened 2016-07-30
Added directory check

marvinl97:
Fixes the error message when doing a fresh install.
e.g.:
```
~$ ./pzserver install
    fetching core_dl.sh...OK
    fetching core_functions.sh...OK
    fetching core_trap.sh...OK
    fetching core_messages.sh...OK
    fetching core_getopt.sh...OK
    fetching command_install.sh...OK
    fetching check.sh...OK
    fetching check_root.sh...OK
    fetching check_permissions.sh...OK
find: ‘/home/pzserver/serverfiles’: No such file or directory
    fetching check_glibc.sh...OK
    fetching info_glibc.sh...OK
    fetching info_distro.sh...OK
    fetching check_tmux.sh...OK
    fetching install_header.sh...OK
```

**Is there a reason to use**
`if [ -n "${functionsdir}" ]; then`
**over**
`if [ -d "${functionsdir}" ]; then`
**?** ([source](https://github.com/dgibbs64/linuxgsm/blob/master/lgsm/functions/check_permissions.sh#L26))

-n just checks if the string isn't empty and -d if the directory exists?

### DIFF
--- a/lgsm/functions/check_permissions.sh
+++ b/lgsm/functions/check_permissions.sh
@@ -9,21 +9,23 @@ local commandname="CHECK"
 local function_selfname="$(basename $(readlink -f "${BASH_SOURCE[0]}"))"
 
 fn_check_ownership(){
-	if [ $(find "${filesdir}" -not -user $(whoami)|wc -l) -ne "0" ]||[ $(find "${rootdir}/${selfname}" -not -user $(whoami)|wc -l) -ne "0" ]; then
-		fn_print_fail_nl "Permissions issues found"
-		fn_script_log_fatal "Permissions issues found"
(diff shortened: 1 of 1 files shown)


UltimateByte:
> -n just checks if the string isn't empty and -d if the directory exists?

Yep.
I have to say i'm still using this table quite often http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html but -d and -n are now assimilated. :D 

For the functions dir, i think it makes no difference to have -n or -d, as if the variable exists and the check_permissions exists, then the folder will exist. However, it makes more senseto me  to have a -d check.
I can only upvote your slight changes. :)

marvinl97:
OK. Found the same table! (:

dgibbs64:
Merged

Issue #212, opened 2014-11-12
[nmrihserver] WARNING: Failed to load 32-bit libtinfo.so.5 or libncurses.so.5.


dgibbs64:
Getting the following error

    WARNING: Failed to load 32-bit libtinfo.so.5 or libncurses.so.5.
      Please install (lib32tinfo5 / ncurses-libs.i686 / equivalent) to enable readline.

dgibbs64:
yum install ncurses
apt-get install lib32tinfo5

dgibbs64:
updated wiki

PR #966, opened 2016-08-03
Added required dependencies to enable readline

marvinl97:
Fixes by adding more required dependencies to CS:S and GMod:
```
WARNING: Failed to load 32-bit libtinfo.so.5 or libncurses.so.5.
  Please install (lib32tinfo5 / ncurses-libs.i686 / equivalent) to enable readline.
```
This enables the access of previous commands with the arrow keys
Like in #212 

**There are maybe more (source) games with this warning.**


dgibbs64:
I have updated the website with the requirements.

PR #987, opened 2016-08-10
Added IPv6 bindings for TeamSpeak 3 server 3.0.13 and newer

Sebi94nbg:
IPv6 support was added to TeamSpeak 3 server version 3.0.13. It's also possible to use IPv6 addresses in the query_ip_whitelist.txt, query_ip_blacklist.txt and tsdns_settings.ini. Please note, that you need to set the IPv6 inside square brackets in the tsdns_settings.ini (and only there) like this: ipv6-ts3.example.com=[2606:2800:220:1:248:1893:25c8:1946]:9987

### DIFF
--- a/TeamSpeak3/cfg/lgsm-default.ini
+++ b/TeamSpeak3/cfg/lgsm-default.ini
@@ -1,11 +1,11 @@
 machine_id=
 default_voice_port=9987
-voice_ip=0.0.0.0
+voice_ip=0.0.0.0, ::
 licensepath=
 filetransfer_port=30033
-filetransfer_ip=0.0.0.0
+filetransfer_ip=0.0.0.0, ::
 query_port=10011
-query_ip=0.0.0.0
+query_ip=0.0.0.0, ::
 query_ip_whitelist=query_ip_whitelist.txt
 query_ip_blacklist=query_ip_blacklist.txt
 dbplugin=ts3db_sqlite3


UltimateByte:
We might need to remove it if it continues causing issues to users.

Sebi94nbg:
Is it maybe possible to add a check, if IPv6 is even configured or not? Then you could decide, if you need to install it with or without IPv6 support.

Every user is allowed to execute the command ``ip address show`` - or short: ``ip a s``. Well... The check could look like this:

```
# ip a s | grep -E "inet6.*scope global"
    inet6 fd00::20c:29ff:feb6:85a8/64 scope global dynamic
```

``scope global`` means, that it's valid everywhere. Count it and if there are one or more such scopes, you can install the IPv6 support based settings file:

```
# ip a s | grep -E "inet6.*scope global" | wc -l
1
```
...else LGSM should install the settings file without IPv6 support.

UltimateByte:
Thanks.
For some reason, somebody came up with
`|bind failed on :::30033; reason: Address family not supported by protocol`
Do you think this is because of the lack of an IPv6 ?

Sebi94nbg:
Yes, because the TeamSpeak 3 server tries to bind the server on IPv4 and IPv6 at startup. Due of that, there isn't any IPv6 address - even a loopback IPv6 address - the bind fails with this message.

The other reason could be, that the kernel does simply not support IPv6, but incase of that, you also wouldn't have any IPv6 address on any NIC and it would also fail.

UltimateByte:
Do poeple really use ipv6 with ts3 ?

Sebi94nbg:
I don't have any statistics about this, but sure. Why not?

Only the hoster of the TS3 servers is able to say, if they're reachable via IPv6 too or not. Much people requested this feature / implementation as you can read [in the official forum thread](http://forum.teamspeak.com/threads/55830-Suggestion-IPv6/).

The new client will always try to connect via IPv6, if the client has a valid IPv6 address. Due of the fact, that IPv6 is faster than IPv4 due of the better header, it's also just better for the clients to use IPv6 instead of IPv4.

Myself is providing TS3 servers with IPv4 and IPv6 support.

UltimateByte:
Best way would be as you said, to detect if the system supports IPv6 or not, and edit the cfg file accordingly before it starts ts3 for the first time and generating the token (otherwise it fails).

However, if some people might like using IPv6, i'm not sure the majority's using it. I personally have an IPv6 and IPv4, but redirect a sub-domain to my IPv4, which allows me to connect onto the server (without specifying a port as it's the only ts3server running so it has the default port, no need of tsdns), and that's it.

In the end my suggestion is : rollback this change until we make the IPv6 detection you suggested ?
If you have a full clever code for that, feel free to PR !

Issue #1001, opened 2016-08-13
[FEATURE] Upload to pastebin with hidden sensitive info


UltimateByte:
```
./gameserver details-share
./gameserver ds
```

Or

```
/gameserver details-support
./gameserver ds
```

Or

```
./gameserver post-details
./gameserver pd
```

Or

```
./gameserver details-post
./gameserver dp
```

How to upload
https://gist.github.com/yevgenko/1803605

`details-share` could use the same function as `details`but with an `if`statement for sensitive info.

UltimateByte:
0:41 - CedarLUG: http://steamcommunity.com/groups/linuxgsm/discussions/0/412448792353031314/#c412448792353224827

cedarlug:
- password
- gslt
- steamuser
- steampass
- pushbullettoken
- authkey
- rcon_password
- rcts_strAdminPassword (SSam)
- sv_password
- zmq_stats_password
- zmq_rcon_password
- pass (Terraria)
- rconServerPassword (Starbound)
- IP Address (this actually comes in really handy for debugging, but could be stripped)
- ???

What else should be sanitized?

UltimateByte:
I think you pretty much nailed it :o))

cedarlug:
[Teaser](http://pastebin.com/71ynisvw)
I believe I have all of the items enumerated above, except IP addresses, able to be stripped.

UltimateByte:
You've coded it ?

cedarlug:
On my fork.

UltimateByte:
Lol, "pd", is a french word for pédéraste, which means basically, "gay".
I'm already laughing thinking about asking people to run ./gameserver pd :o))

Good job btw..

Something cool would be to mask off the user only if it's something different than anonymous. That way we'd know if the user tries to login to steam while it's unnecessary.

Does it mask off info from fn_parms already ? 
Did you try pastebin upload ?

cedarlug:
Good progress  on this - need some input though.

I have it so that steamuser=anonymous and steampass empty are reported rather than stripped.

What would you like to see for the scenario where the rustserver has default CHANGE_ME password?

Currently there's a warning output in the details output.  Should that be reported in the details that we ask users to post, or stripped because it's a default password used for rcon on someone's server?

cedarlug:
... and yes, I've worked the curl to pastebin in.

UltimateByte:
[...]

PR #1037, opened 2016-08-28
Post details to pastebin - stripped of sensitive info

cedarlug:
First pass at #1001 

Basic gist:

command_details.sh already contains all of the functions to pull and parse the gameserver info.  To leverage the already-defined functions and to keep things sane going forward, I changed command_details.sh to only output details when the POSTDATA variable is undefined.  That allows command_postdetails.sh to SOURCE command_details.sh to define all of its functions.

In general, "postdetails" is just calling "details," grabs all of the information, sed's it six ways from Sunday, strips off colorizing and console control codes, and then posts the results to pastebin.


dgibbs64:
Pop this in to the development branch please :)

dgibbs64:
IN face yeah just pop this in to its own branch called pastebin then I can do full tests and make any changes :D. Good work!!

cedarlug:
Thanks for taking a look.  Moving it to pastebin.

PR #1044, opened 2016-08-30
adding Quake 2

hurdad:


hurdad:
Ahh I think issue is that file name can't have hypen. Have to convert it to underscore.  E.g + exec quake2_server.cfg

PR #1046, opened 2016-08-31
adding Quake 3

hurdad:


dgibbs64:
@hurdad I need clarification is this Quake 3: Arena or Quake 3: Team Arena

hurdad:
I believe the bundle I created includes 'Team Arena' expansion. 
So, I don't know how you want to name it. 
The original game is 'Quake III Arena', the expansion is ' Quake III: Team Arena'

dgibbs64:
I am happy to report it appears to work. However I am confused about a few things. Mainly because there are no good guides on Quake 3 arena servers. 

I have tried out Quake 3 Arena and Team Arena. Team Arena appears to not have any active servers so I think that is probably not used at all?

I installed qtracker to get more info on quake 3 servers most use build Q3 1.32c linux-i386 May  8 2009 rather than 1.32b. Or they use ioq3 https://github.com/ioquake/ioq3

Do we need to update the download for 1.32c?

hurdad:
Glad it worked. Not sure why I used 1.32b over 1.32c. When I open Q3 in steam it just says 1.32, and it connected ok to the 1.32b server.

I can attempt to upgrade the bundle to 1.32c and test if you want.

dgibbs64:
Yes please. once 1.32c is updated then it should be ready :).

hurdad:
I tested 1.32c and it is working. please proceed with merge.

dgibbs64:
Released ! https://gameservermanagers.com/lgsm/q3server/

PR #1073, opened 2016-09-14
Add variable to set email's sender 'emailfrom'

dgibbs64:


UltimateByte:
Nice. Need to add it to other games though... pfiouh

Issue #1059, opened 2016-09-07
Mods Loading in 'debug'-mode, but not with 'start'


curiousfREAk:
Following Issue:
If i start the Server with "arma3server debug ("takes 1'15'' until connected to steam servers), the Mods are loading.
Starting it up with "arma3server start" only the first mod is loading .....
The mods-Parameter didn't work in with the Syntax from the the Example ( mods/\@CBA_A3\;mods/\@task_force_radio ...)
My Mod-Line Syntax is:
mods="mods/@cup_terrains;mods/@ace;mods/@alive;mods/@burnes..."
 serverfiles/mods/\* is 775 for the arma3server user ...

.... anyone else?

best regards

fREAk

UltimateByte:
Hi,

I've seen topics/issues about people having the syntax working either in debug or normal start.
This needs investigation, maybe we could fix it with some sed && || some good plays with single and double quotes.
However, the most important for you would be to have it working in normal mode for now.
Relevant things that might help : 
https://github.com/GameServerManagers/LinuxGSM/issues/739
https://github.com/GameServerManagers/LinuxGSM/issues/651

cedarlug:
I'll take a crack at figuring out what's going on.

cedarlug:
@curiousfREAk  Would you please enumerate a full list of your mods?  I'd like to test with everything that you're currently using.

curiousfREAk:
This was faulty diagnosed by, sorry.
So here we are now: Once i load one mod with :: mods="mods/@cba_a3" the mod is loaded correctly, ...
DEBUG => `
13:45:42                                               name |               modDir |    default |               origin |                                     hash | hashShort | fullPath
13:45:42 ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
13:45:42                       Community Base Addons v3.0.0 |              @cba_a3 |      false |             GAME DIR | 5e4d58da7d3525e9089577880a322b2e9ccf1767 |  dc1227b1 | /home/arma3server/serverfiles/mods/@cba_a3
13:45:42                                  Arma 3 DLC Bundle |            dlcbundle |       true |            NOT FOUND |                                          |           |
13:45:42                                        Arma 3 Apex |            expansion |       true |             GAME DIR | 3b262d55f1bf131ac7ba7e8f52db442675e78deb |  734697dd | /home/arma3server/serverfiles/expansion
[...]

Issue #739, opened 2016-03-04
Arma3Server START vs DEBUG error


Talven81:
When issuing ./arma3server START the following mod variable works.
mods="@CBA_A3\;@ace\;"

When issuing DEBUG the escape backslashes cause the mods to fail to load and only works if as such.
mods="@CBA_A3;@ace;"

This is due to direct issuing of the command in DEBUG vs. issuing the command during START through tmux requiring escape backslashes for the string to be accepted.

This can be fixed by modifying command_debug.sh and adding in on line 56 a translation string to strip out the backslashes.
parms=$(echo $parms | tr -d \\)

This also applies to servermods variable, but the above takes care of both.

igeighty:
i have just noticed this also. in compiling a tutorial for the mods folders, i was double checking and came across this issue, @Talven81 i will test it out under both debug and standard.
@UltimateByte if it pans out i will do a pull request for it. Testing it out tonight.

Thanks @Talven81 i think you have saved a lot of tooth pulling :)

igeighty:
Nope I can't make it work, @Talven81 could you post your modified command_debug.sh to pastern please mate ?

dgibbs64:
Is there an update to this? Or can it be closed?

igeighty:
Hi Dan,
@Talven81 never posted back his solution,
indeed there is a difference in how debug and start works. I have got nowhere with this though as i cannot code tmux, it is well above me.

for now i have been just working around this issue. i am more than happy to test any ideas on it you may have.

i did notice that
`parms=$(echo $parms | tr -d \)` threw errors, yet `parms=$(echo $parms | tr -d \ )` (with space before trailing bracket got further. but errored.

in #651 i posted a few example of what works and what doesn't. in all honesty though i stopped hosting mods and just went vanilla.

regards.
Doug

cedarlug:
https://github.com/GameServerManagers/LinuxGSM/wiki/Arma-3

All works as it should. Confirmed with @XxROBOTICSxX today, walking him through the process.

Issue #651, opened 2016-01-27
arma 3 @mods not working in linux game server manager


BlackhatEspeed:
I've been having some issues getting @mods to load from the arma3server.cfg file that's outside of the serverfiles directory on my box, i've got all the mods uploaded , I tried putting them in the serverfiles directory and I tried putting them in their own /mods/ folder inside of the serverfiles directory and I cant get the mods to load for the life of me.

UltimateByte:
Hi there,

I've never played around with Aram 3 servers yet, but it could happen soon. Do you think it's an LGSM script issue ? If not, please post there : https://steamcommunity.com/groups/linuxgsm/discussions

flexgrip:
This is weird. Sometimes I can get the debug screen to list the mods but it doesn't like them.

Anyone have more info on this? I assumed the -mod"mods/@something" string has to be just right.

acidburn101:
I gave up using it i just added the mods to the startup param directly. 

```
fn_parms(){
parms="-ip=${ip} -name=SERVERNAME -cfg=${networkcfgfullpath} -config=${servercfgfullpath} -mod=@CBA_A3\;@AnotherMod"
}
```

Mods must be under serverfiles\

flexgrip:
Yeah. I noticed the same thing. They must be in the same root level as the arma server executable. Which is very weird. I guess we're spoiled linux users...

That being said, I gave up on this altogether. This script works great when you want to run a vanilla arma server. However, if you want to run anything more than a mod or two, **arma 3 server on linux is useless**. Segfaults, half the mods don't work, random crashes. Then when you finally do get the damned thing to run, all of the clients trying to connect get memory access errors. I had better luck getting the windows version of arma3server to run via wine!!!

Sometimes I wish Bohemia would go into their office, one day, and throw away any PC's they have. That way they are forced to develop arma under a better environment like linux or osx and bring their client/server platforms into the modern age. And I guarantee it would run more efficiently as well.

Sorry for the rant. A guy can dream...

igeighty:
It is the commented lines in the arma3server script that give example paths that lead me down the wrong path.

`# ARMA 3 Modules`
`# add mods with relative paths:`
`# mods/\@CBA_A3\;`
`# or several mods as:`
`# mods/\@CBA_A3\;mods/\@task_force_radio`

When i used this example syntax i was unable to get any mods to load.

[...]

PR #1077, opened 2016-09-18
ARMA3: Fixing tmux/debug module discrepancies

cedarlug:
Fixes: #1059, #739, #651 
The gist of the solution is to adhere to the escaping of just semicolons in the script, and using pure semicolons in debug mode.

Comments in the server script also updated to reflect the actual usage paradigm.


UltimateByte:
Amazing !

PR #1081, opened 2016-09-19
Fast improve for more automatization capabilities

Smile42RU:
If you need use LGSM as automatically installer you should edit (for example) arkserver file. Not too automated, uncomfortably.

And after operations on the server credentials from Steam account stay in file — it not secure, i think.

I see another way — just use environment variables for steamuser && steampass like as:
"globalsteamuser=user globalsteampass=pass ./sbserver install"

You can run install, start or stop operations without any trace of yours Steam credentials, of course if you not logging console history or cleaning up it by simple command "history -c".

Also, if you run this command from another server (for example, automation controller server) any changes in files and any cleaning not required — it's awesome.

### DIFF
--- a/lgsm/functions/check_steamcmd.sh
+++ b/lgsm/functions/check_steamcmd.sh
@@ -16,7 +16,14 @@ fn_install_steamcmd(){
 }
 
 fn_check_steamcmd_user(){
-	# Checks if steamuser is setup.
+	# Check if steamuser is setup in ENV.
+	if [ -n "${globalsteamuser}" ]; then
+		fn_print_ok_eol_nl
+		fn_print_info "Steam Credentials taken from environment variables"
+		steamuser="${globalsteamuser}"
+		steampass="${globalsteampass}"
+	fi
+	# Checks if steamuser is setup in config file.
 	if [ "${steamuser}" == "username" ]; then
 		fn_print_fail_nl "Steam login not set. Update steamuser in ${selfname}."
 		echo "	* Change steamuser=\"username\" to a valid steam login."


Smile42RU:
Please, give me a feedback asap for next step.

dgibbs64:
Good idea. I like it. It would be good if it was possible to use steamuser= and steampassword= rather than have to use globalsteamuser=. However I am not sure how we would go about that. I will not be merging this just yet however this is something we can look at in the future :D. Thanks for the idea!!

Smile42RU:
@dgibbs64 Daniel! Thank you.

Tell me all of yours concerns and let's try solve it, because we use this approach in our project and we want implement it to master branch to get rid supporting our custom repo with custom builders.

If you have a time, than i suggest talk about this feature in details here or in any IM for greater efficiency.

I and my team have all competence and enough experience for develop any your vision about this feature, but we need get detailed description as soon as possible. 

We really want help project which we using in our products and we can solve not only specified for us tasks, but we can make contribution in most pieces project. For example: my team has translate most of the Beego framework documentation from Chinese and English to Russian (https://github.com/beego/beedoc, you can find my commits as confirmation) and we've improve many projects on GitHub related with Golang and PHP.

I'm sorry that so hurry, your product is great, it solves many of our problems, but this functionality is really important for us because it will allow us to move forward.

Smile42RU:
@dgibbs64 Also a little thought about "global" prefix for steamuser / steampass variables. Yep, it can works fine without any prefixes, but I did it on purpose, because environment variables can be used with "export" linux command. I mean:

export globalsteamuser=user
export globalsteampass=pass
./arkserver start
-- some time later --
./arkserver start

We do not do so, but it works and I thought that it is necessary to separate the variables for reduce chance misreading.

Anyway I'll be happy to hear your opinion.

UltimateByte:
Hey, 
One word then i go to sleep :)
Why do you need this ? Might give ideas on how to implement it.
Because from my perspective, use chmod o-rwx game server, and this is enough safety.
Also, I don't know for ark, but for many games, you just need to be logged in as a steam user, not to actually own the game, so you can make spare accounts for that matter.

Hopefully we can find the best solution to suit everyone.

Smile42RU:
@UltimateByte "use chmod o-rwx game server, and this is enough safety"... Yes, it works if it is only yours server :)

We are game hosting. We massively using LGSM for control many games on user's VPS and we can't persistently provide our credentials into user VPS, because user have full root access to server.

Okay, we can create some fake/spare accounts, but we can't use it in LGSM with enabled Steam Guard, because in that case we can forget about automatically operations. And anyone who can read "arkserver/***server" (our client) can get access to fake/spare account. Also we can't approve any solutions with creating many fake accounts :) It not works with Valve / Steam agreements.

It is not the best solution, unfortunately, sorry.

We are spent many hours for investigating this question and we have only one comfortable and safety approach to provide credentials into LGSM.

JimTR:
Ok why not produce a single user on each VPS (I guess they are remote to your hosting station)  give them root privs , no login and then run some code that takes into account the user and processes the data

Smile42RU:
@JimTR I can't understand what you try offer, sorry.

cedarlug:
If I understand @Smile42RU's business model (or a reasonable approximation of it), they are selling the game server as a service, and not shell access.  They deploy the server out-of-band and give rcon and maybe ftp to the end user.  But  the automation of the gameserver deployment is part of what is being sold as a service.  In this way, the end user doesn't receive steam credentials for their game, and they would like to keep those credentials from being disclosed in the script or otherwise found in the user's logs (.bash_history, for example).

Am I close, @Smile42RU ?

JimTR:
Whatever

Smile42RU:
@cedarlug Partly. Actually we provide real KVM VPS (VDS), but other things was described pretty correctly.

We strongly concerned about security aspect of open use Steam credentials for LGSM operations as it described in docs.

BTW, @UltimateByte, we investigated 46 games from LGSM supported list and 35 of them using anonymous auth, 6 can be operated by fake/spare account and 4 required a valid license. A few others which be excluded from the research, because not used Steam or games too old.

At first sight it looks like not reasonable for change/improve credentials management in LGSM, but this 10 games that use non-anonymous Steam account including a great titles like as ARMA3, 7DaysToDie, Natural Selection 2, Starbound and Terraria. It's ones of most popular games among the players.

cedarlug:
I (personally) don't think that check_steamcmd.sh is the right place for this check.

Instead, why not just override the assignment at the top level and use the already-defined `steamuser` and `steampass` variables?  Based on what you're currently doing, what your goal is, and the code changes in the PR above, this looks to be a viable alternative for what you want to accomplish.   What you are looking to do doesn't require LGSM integration.

E.g.:
Do a full git clone and enter the LinuxGSM directory.

Then issue:
```bash
sloppy@delli7:~/LinuxGSM$ for x in $(find . -maxdepth 2 -mindepth 2  -type f  | grep -E -v ./test\|./.git\|./appmanifest\|./functions)
> do
> sed -i "s/^steamuser=\"username/steamuser=\"\${steamuser=username}/g" $x 
> sed -i "s/^steampass=\"password/steampass=\"\${steampass=password}/g" $x
> done
sloppy@delli7:~/LinuxGSM$
```

Afterward, all steam creds are set up across *all* scripts in a way that make them defer to the environmental setting if present.  They should all look similar to this:
```bash
sloppy@delli7:~/LinuxGSM$ grep ^steam Terraria/terrariaserver 
steamuser="${steamuser=username}"
steampass="${steampass=password}"
sloppy@delli7:~/LinuxGSM$ 
```

And voila!  You can now override all steamuser/steampass settings when running the scripts:
```bash
slppy@delli7:~/LinuxGSM/Terraria$ username=cedarlug userpass=NotMyPassword ./terrariaserver ai
```
For any variable not set in the environment, the behavior defaults to the usual assignment.

If you pull the scripts using wget, the same two sed commands above would post-process the gameserver script in the same way (you won't need the "find" command).

Good luck.

(I don't think this should be the default in LGSM either, but would lobby for this approach over the `check_steamcmd.sh` approach.)

Smile42RU:
@cedarlug 

Thank you for feedback and snippets, they looks nicely. 

I agree that check_steamcmd.sh is not the best, but it's really «fast improve» without deep dive. I'm sure that you and team will can find more appropriate place.

Well, also i see a few things and cases which concern me now.

1. About sed: It will be work only if client do not touched any config earlier than our "scripts" replace user/pass stuff. It is unlikely that this will happen, but still. I don't think that justified assumption thinking that it will never happen.

2. I predict that client will erase ${steamuser=username} in config file because "=username" is too difficult for most not tech guys and kids. And, for example, if client change steamuser or steampass in ***server file then «environment variables» in your case will be broken. 

It's very tender spot and i think with this approach we (and any other guys and teams they want automate something) will often meet different tickets from clients related with this feature.

3. As i see it not will works with anonymous user by default. It's still a "patch" for manually use and not looks like as something that will work in all conditions and automatically.

Please, correct me if I'm wrong.

cedarlug:
@Smile42RU  I understand your concerns.  But I believe post-processing on your part is still the way it should be.  I don't like the check_steamcmd.sh solution for a few reasons: 

1. It puts a workaround solution "under the hood" in support of a specific project rather than addressing the general goals of the LGSM project.  If this approach were to be a solution, it should be universally available to all (not just a buried one-off solution) - meaning that your clients would still be able to bypass it.  **And indeed they still can!** (See related point 3. below, too.)
2. Users will always break code.  Always.  If you really want to assure that users don't modify the contents of the script, then put it in your TOS and simply "chatter +i" the file.  This is an especially-easy solution if you're using openVZ/LXC/Docker containers for your deployment and can effect change from outside the containers on the physical host.
3. If any of your customers were to post questions on any of the various LGSM forums, asking how to specify or override steam credentials, we'd certainly tell them anyway. 
4. For any gameserver that you deploy, you're likely already scripting the "wget" or a "cp" to pull the gameserver and "chmod +x" the script to deploy a container/VPS.  In my opinion, adding a sed afterward (the 2 sed commands are easily combined into a single command) should be a trivial amendment to what you're already automating..

My biggest concern is the 1st: You're asking to bury a workaround in the LGSM plumbing.

cedarlug:
Also with regard to your point 3:  I don't know if it's a language issue, but your comment implies that you don't understand that anonymous steam credentials are still completely supported.  The result in *all* cases is the same behavior as the original, unmodified gameserver script.

Smile42RU:
@cedarlug
@dgibbs64 
@UltimateByte 

Guys, thank you for productive conversation. Well, I think that you need something like this functionality, but may in another embodiment.

Sadly, but i see that we need support our fork and stop spend time for discussions :)

Issue #1143, opened 2016-10-20
cached is now missing in details


dgibbs64:
It looks like the free command has changed as the cached option is now missing. I am not sure why this is. 
```
Mem:       total   used   free   cached
Physical:  992M    492M   100M
Swap:      0B      0B     0B
```

A typical output of free -h now read like so

```
lgsm@ubuntu-1604-lgsm-dev:~/ns2server$ free -h
              total        used        free      shared  buff/cache   available
Mem:           992M        490M         96M        3.1M        405M        344M
Swap:            0B          0B          0B
```

I can simply remove the cached option or look at adding the buff/cache and available option

UltimateByte:
Got the right output with Debian.
Need to find out an universal way.

```
Mem:       total   used   free     cached
Physical:  31G     29G    1.7G     19G
Swap:      1.0G    292M   729M
```

dgibbs64:
fixed by @UltimateByte

UltimateByte:
@dgibbs64 Without a merge it's not fixed :)

UltimateByte:
https://github.com/GameServerManagers/LinuxGSM/pull/1144

Issue #1142, opened 2016-10-20
[config] file should be added from default if not found upon start


UltimateByte:
Most games will be useless without a config file, so if one is available, it should be copied upon server start. Would make the process of duplicating a server with a different servicename easier as well.
```
if [ ! -f "${servercfgfullpath}" ]; then
# install a default config file
fi
```

UltimateByte:
Fixed on next merge

Issue #1141, opened 2016-10-20
[Rust] Config file not created 


UltimateByte:
Probably due to the config update, rust config isn't created after a successful install.

```
copying server.cfg config file.
'/home/rust/lgsm/default-configs/server.cfg' -> '/home/rust/serverfiles/server/rust-server/cfg/server.cfg'
cp: cannot create regular file '/home/rust/serverfiles/server/rust-server/cfg/server.cfg': No such file or directory
changing hostname.
sed: can't read /home/rust/serverfiles/server/rust-server/cfg/server.cfg: No such file or directory
changing rcon/admin password.
sed: can't read /home/rust/serverfiles/server/rust-server/cfg/server.cfg: No such file or directory
```

I think this is due to rust-server/cfg not existing until the server is started once. 
Will investigate.

UltimateByte:
Fixed on next merge.

Issue #1140, opened 2016-10-20
[backup] error upon anything if backup folder is empty


UltimateByte:
```
ultimatebyte@game:~$ ./mumbleserver stop
[ .... ] Stopping mumble-server: find: `/home/ultimatebyte/backups/*.tar.gz': No such file or directory
ls: cannot access /home/ultimatebyte/backups/*.tar.gz: No such file or directory
date: : No such file or directory
du: invalid zero-length file name
[ INFO ] Stopping mumble-server: Mumble Port 64739 is already stopped
```

This probably happens because a check is assuming that there will always be backups in the backups folder. Need to find out which one. :)

UltimateByte:
Guilty 
https://github.com/GameServerManagers/LinuxGSM/blob/ultimatebyte-dev/lgsm/functions/info_distro.sh#L106

UltimateByte:
Will be fixed on merge.

PR #1144, opened 2016-10-21
fixes and improved backup

### LINKED ISSUE #1098: backupsdays backup autoclear
It won't be very hard to implement and will allow people to run it as a cronjob.
We can use this as a template : https://github.com/GameServerManagers/LinuxGSM/blob/master/lgsm/functions/logs.sh#L49

Must-have would be to check if a backup is already pending before starting another, in case somebody messes up and runs the cronjob every minute instead of every day for example. 

Settings the amount of backups to keep would be added to the main script with a backupdays variable, and a value of "0" would mean "infinite". 

I'll probably code a bit in the next days or weeks, so i'm assigning myself.

UltimateByte:
I'm satisfied from my tests, requires some double check & testing and it'll be fine.

Fixes
#1098
#1142
#1141
#1140
#1137
edit: and #1143


dgibbs64:
Is this backwards compatible?
```
maxbackups="4"
maxbackupdays="30"
stoponbackup="on"
```
Ensure that these commands have a default should they be missing. lookin core_functions.sh under 

```
## Code/functions for legacy servers
```

UltimateByte:
@dgibbs64 If variables don't exist, there is no default values set to not annoy people that don't want backup autoclear, so the backup rotation will just not happen.
That's why i start with `if [ -n "${maxbackupdays}" ]&&[ -n "${maxbackups}" ]; then` to make sure that the process will go flawlessly.

Last commit should fix #1143 please test on ubuntu to confirm.

dgibbs64:
@UltimateByte 

Its looking good so far. Im just looking at your code to make sure I understand it. Came across this error though.
```
[  OK  ] Backup jc2-server: Backup created: jc2-server-2016-10-22-152343.tar.gz is 30M size
find: invalid argument `+' to `-mtime'
```

UltimateByte:
@dgibbs64 Ubuntu ? 😷 
Don't have this issue. o_O
I guess i should move -type f before -mtime to solve this on ubuntu

UltimateByte:
Is it solved now ?

dgibbs64:
Yup fixed

dgibbs64:
A small change. The backup vars should not be listed under advanced vars since we want the user to feel comfortable to change them. Like so.

https://gist.github.com/dgibbs64/a2bc9226763b4112482ee19ff7d070c7

Also the wiki page will need to be updated to users know what exactly they are doing with eh new backup system

UltimateByte:
@dgibbs64 I'll update the wiki after merge.
logdays should come close to backups as well then, what do you think ?

PR #1156, opened 2016-10-25
Fixes #1154

### LINKED ISSUE #1154: LGSM recently blocks nested sessions
I run multiple LGSM servers on the same machine. I create a tmux session for each user and login with that user account and run LGSM as a nested session.

I've been doing this for more than a year now and it has worked flawlessly. I installed a new server on 2016-10-13 with no problem. When I did the same today, I got the following errors: 

[ FAIL ] Install csgo-server: tmuxception error: Sorry Cobb you cannot start a tmux session inside of a screen session.
Information! LGSM creates a tmux session when starting the server.

Tmux allows nested sessions.

CasperAlant:
Users are prevented from starting an LGSM server within a tmux session, unless the session was started by another user.

### DIFF
--- a/lgsm/functions/check_tmuxception.sh
+++ b/lgsm/functions/check_tmuxception.sh
@@ -17,16 +17,5 @@ fn_check_is_in_tmux(){
 		core_exit.sh
 	fi
 }
-fn_check_is_in_screen(){
-	if [ "$TERM" == "screen" ];then
-		fn_print_fail_nl "tmuxception error: Sorry Cobb you cannot start a tmux session inside of a screen session."
-		fn_script_log_fatal "tmuxception error: Attempted to start a tmux session inside of a screen session."
-		fn_print_information_nl "LGSM creates a tmux session when starting the server."
-		echo "It is not possible to run a tmux session inside screen session"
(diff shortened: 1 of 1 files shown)


dgibbs64:
This looks like a function has just been removed. There needs to also be a method for confirming that the user is not trying to use screen either.

UltimateByte:
Better :p 
Maybe `if [ -z "${TMUX}" ]&&[ -n "${STY}" `] ?

CasperAlant:
${TMUX} is unset after logging in to the lgsm account. The problem was that ${TERM} gets set to "screen" and stays that way even after logging into the lgsm account. By simply checking for a screen session using the ${STY} variable, my problem is solved.

UltimateByte:
Well, checking ${STY} seems like a good option at first glance then. :)
What does ${STY} variable points to exactly ?

CasperAlant:
As per the screen manual, ${STY} holds information about the screen window manager or the "Alternate socket name." Screen checks it when a new screen is created and will make a new window instead of a nested screen if it is set.

UltimateByte:
I'll patch it using the new branch, no worries :)

Issue #1221, opened 2017-01-03
[Server Request] Multi Theft Auto


ChaosMTA:
Would be nice to have MTA implemented if SAMP is being considered. MTA is another popular modification for Grand Theft Auto San Andreas.

I run a popular server on there and would be willing to help test or contribute. I am not good enough in bash to maybe be that much help.
This is a bash script I made to deploy a MTA server using the default resources or use your own Git repository as the source. Not sure if it would be any help to you. (For Debian)

```bash
#!/bin/bash

HELP="Usage:

\t[-g GIT_REPO]
\t[-p DB_PASS]
\t[-h HELP]\n"

BINARY="https://linux.mtasa.com/dl/152/multitheftauto_linux_x64-1.5.2.tar.gz"
CONFIG="https://linux.mtasa.com/dl/152/baseconfig-1.5.2.tar.gz"
RESOURCES="https://github.com/multitheftauto/mtasa-resources/archive/master.tar.gz"

while getopts ":a:g:p:h" arg; do
  case $arg in
    g) #g allows the user to initiate a git repository in the deathmatch folder
		GIT_REPO="$OPTARG"
		;;
    h) #h indicates that the user requests to see this script's usage
		printf "$HELP"
		exit
		;;
		p) #p allows the setting of the mysql root password
		DB_PASS="$OPTARG"
		;;
	*)
		printf "$HELP"
		exit -1
		;;
  esac
done

# make sure server is up to date
apt-get -y update && apt-get -y upgrade

# gather binaries
wget -O mta.tar.gz $BINARY
# extract mta server into mta directory
mkdir /opt/mta/
tar xzvf mta.tar.gz -C /opt/mta/ --strip-components=1

# load in the mysql module
wget https://nightly.mtasa.com/files/modules/64/mta_mysql.so
mkdir /opt/mta/x64/modules
mv mta_mysql.so /opt/mta/x64/modules

# fix mysql error
wget https://nightly.mtasa.com/files/modules/64/libmysqlclient.so.16
mv libmysqlclient.so.16 /usr/lib

# cleanup
rm -f mta.tar.gz

# install MariaDB
apt-get -y install software-properties-common
apt-key adv --recv-keys --keyserver keyserver.ubuntu.com 0xcbcb082a1bb943db
add-apt-repository 'deb [arch=amd64,i386] http://nyc2.mirrors.digitalocean.com/mariadb/repo/10.1/debian jessie main'
apt-get update

[...]

PR #1224, opened 2017-01-03
Adding Multi Theft Auto Support

ChaosMTA:
This is my first pull request ever on Github and I'm not the most experienced in bash so bare with me..

This commit aims to add support for Multi Theft Auto. MTA requires three things for it operate:

- binary files
- configuration files
- resources(server script that the binaries interpret)

I created a file called `install_mta_resources.sh` which will download the required libmysqlclient socket in order to use [database functions](https://wiki.multitheftauto.com/wiki/Server_Scripting_Functions#SQL_functions) on the server and will download the default resources for use.

I'm not sure how to test this code since I'm not very familiar with it but I will if someone points me in the right direction. The only thing that is missing is the configuration files found [here](https://linux.mtasa.com/dl/153/baseconfig-1.5.3.tar.gz) need to be added to the Game-Server-Configs repository. Pull request: https://github.com/GameServerManagers/Game-Server-Configs/pull/5

Please let me know if there is anything I can do to improve it.


UltimateByte:
Good job man.
Did the wiki help ? Because it looks good at first glance.
Did you also get details/monitor/update functions to work ?

The only thing wrong i see is the look of the main script that doesn't fit recent changes about the look and new settings available. Like this one for example: https://github.com/GameServerManagers/LinuxGSM/blob/master/UnrealTournament99/ut99server
Also, the version should now be the date in the form YYMMDD.

You can make changes in your branch if you will, they will be committed to this pull request.

ChaosMTA:
Done updating the version number and look of the file to match the others.

I used the wiki and found it helpful yes, although I recommend adding something about the formatting for the version number since I couldn't find that.

As I said in my original pull request:

> I'm not sure how to test this code since I'm not very familiar with it but I will if someone points me in the right direction. 

Perhaps adding a testing/debugging page on the wiki would also help me out in that regard. So I'm not 100% sure if the monitor functions are working, based on the pull request for SAMP, it works so it should work here too. But if not I will make the needed changes.

Thanks for the feedback @UltimateByte

UltimateByte:
I'm glad it actually helps ! @dgibbs64 helped me directly when i was beginning, my goal with this page is to help other devs the same way and even more efficiently.

I added some more information. Since you're probably the first one using it,, don't hesitate to notify me for missing information. 
https://github.com/GameServerManagers/LinuxGSM/wiki/Developing-LGSM#commands
https://github.com/GameServerManagers/LinuxGSM/wiki/Developing-LGSM#versioning
https://github.com/GameServerManagers/LinuxGSM/wiki/Developing-LGSM#testing-and-debugging-your-code

ChaosMTA:
This is ready to merge, sorry about the commit spam.

Everything it working properly now though. Something that should be noted is my change to core-dl. This is the only game server to use tar.gz so I don't see it affecting anything negatively. But strip components has to be used in order to extract properly.

lgsm/functions/core_dl.sh
```diff
@@ -58,7 +58,7 @@ fn_dl_extract(){
 	mime=$(file -b --mime-type "${filedir}/${filename}")
 
 	if [ "${mime}" == "application/gzip" ]||[ "${mime}" == "application/x-gzip" ]; then
-		tarcmd=$(tar -zxf "${filedir}/${filename}" -C "${extractdir}")
+		tarcmd=$(tar -zxf "${filedir}/${filename}" -C "${extractdir}" --strip-components=1)
```
ChaosMTA: Note this change

dgibbs64:
Good effort with this!! I will look at integrating this with LinuxGSM. Just bare with it though will take me some time :)

ChaosMTA:
@dgibbs64 Any idea how long? Or can I help in anyway with getting it integrated?

lgsm/functions/core_dl.sh
dgibbs64: Never come across --strip-components=1 before. what does it do and could it affect other extractions?

lgsm/functions/install_server_files.sh
```diff
@@ -39,6 +39,8 @@ fn_install_server_files(){
 		fileurl="http://files.gameservermanagers.com/UnrealTournament3/UT3-linux-server-2.1.tar.bz2"; filedir="${tmpdir}"; filename="UT3-linux-server-2.1.tar.bz2";  executecmd="noexecute" run="norun"; force="noforce"; md5="2527437b46f1b47f20228d27d72395a6"
 	elif [ "${gamename}" == "Wolfenstein: Enemy Territory" ]; then
 		fileurl="http://files.gameservermanagers.com/WolfensteinEnemyTerritory/enemy-territory.260b.tar.bz2"; filedir="${tmpdir}"; filename="enemy-territory.260b.tar.bz2";  executecmd="noexecute" run="norun"; force="noforce"; md5="f833f514bfcdd46b42c111f83350c5a7"
+	elif [ "${gamename}" == "Multi Theft Auto" ]; then
+		fileurl="https://linux.mtasa.com/dl/153/multitheftauto_linux_x64-1.5.3.tar.gz"; filedir="${tmpdir}"; filename="multitheftauto_linux_x64-1.5.3.tar.gz"; executecmd="noexecute" run="norun"; force="noforce"; md5="77caf91fe280877a8d21f8046d5f42ba"
```
dgibbs64: Do MTA release updates? If so an update feature may need to be developed.

lgsm/functions/install_mta_resources.sh
```diff
@@ -0,0 +1,58 @@
...
+	echo "================================="
+	sleep 1
+	if [ ! -f /usr/lib/libmysqlclient.so.16 ]; then
+		fn_print_warn_nl "libmysqlclient16 not installed. Installing.."
+		sleep 1
+		sudo -v > /dev/null 2>&1
+		if [ $? -eq 0 ]; then
+    	fileurl="https://nightly.mtasa.com/files/modules/64/libmysqlclient.so.16"; filedir="${tmpdir}"; filename="libmysqlclient.so.16"; executecmd="executecmd" run="norun"; force="noforce"; md5="6c188e0f8fb5d7a29f4bc413b9fed6c2"
+    	fn_fetch_file "${fileurl}" "${filedir}" "${filename}" "${executecmd}" "${run}" "${force}" "${md5}"
+			sudo mv ${tmpdir}/${filename} /usr/lib/${filename}
```
dgibbs64: This needs to change. libmysqlclient16 needs to be installed as a dependency (detect_deps.sh) using apt or yum. I am against LinuxGSM making any changes to the core OS like this as its a back "hacky". Alternatively a glibcfix type of method may be required.

lgsm/functions/install_mta_resources.sh
```diff
@@ -0,0 +1,58 @@
...
+	fn_dl_extract "${filedir}" "${filename}" "${resourcesdir}"
+  echo "Default Resources Installed."
+}
+
+fn_install_libmysqlclient16
+
+if [ -z "${autoinstall}" ]; then
+	echo ""
+	while true; do
+		read -e -i "y" -p "Do you want to have the default resources downloaded? (Server is inoperable without resources!) [y/N]" yn
```
dgibbs64: I think this question is not required as the server must have resources to run. The answer will always have to be yes. Otherwise there is no point installing the server. 

I think that this download should be integrated in to install_server_files.sh next to multitheftauto_linux_x64-1.5.3.tar.gz

lgsm/functions/info_config.sh
```diff
@@ -576,6 +576,38 @@ fn_info_config_sdtd(){
...
+		else
+			ase="Disabled"
+		fi
+
+		# Not Set
+		port=${port:-"NOT SET - Defaults to 22003"}
+		httpport=${httpport:-"NOT SET - Defaults to 22005"}
+		ase=${ase:-"NOT SET - Defaults to Disabled"}
+		servername=${servername:-"NOT SET"}
+		serverpassword=${serverpassword:-"NOT SET - Defaults to none"}
```
dgibbs64: No requirement to state Defaults to none or Defaults to 22005. The default config will have these by default. These settings should not be missing from the config even if they have a default.

lgsm/functions/info_glibc.sh
```diff
@@ -131,6 +131,9 @@ elif [ "${engine}" == "refractor" ]; then
 elif [ "${gamename}" == "Wolfenstein: Enemy Territory" ]; then
 	glibcrequired="2.2.4"
 	glibcfix="no"
+elif [ "${gamename}" == "Multi Theft Auto" ]; then
+	glibcrequired="NOT REQUIRED"
```
dgibbs64: Please double check and confirm this for me using "detect-glibc" option in LinuxGSM.

dgibbs64:
I do most of the work in to master. Its mainly just me checking everything meets the right standard before released and ensuring everything is OK. I have a large backlog of stuff currently.

I have left a load of comments of stuff that needs looking at. Please review them and make any changes required. 

Overall im very impressed with how well you have done with this. Well done!

dgibbs64:
Just found this might be useful for you https://wiki.multitheftauto.com/wiki/Installing_and_Running_MTASA_Server_on_GNU_Linux#Automatic_installer

lgsm/functions/core_dl.sh
ChaosMTA: --strip-components=NUMBER
       strip NUMBER leading components from file names on extraction

Basically it removes the leading, empty directory from the tar file. In which MTA has.

It depends on the other extractions. However I'm not sure how else to handle this while still using your extract function since I can't pass arguments to it. If it's acceptable to not use your extract function then I can do it differently.

lgsm/functions/install_server_files.sh
ChaosMTA: They do release updates however an update feature would mean being able to check the server version from bash, not currently something I believe it possible but I can try to find out more. Regardless I'm not sure how an updater would work because the link would change for the update it needs.

lgsm/functions/install_mta_resources.sh
ChaosMTA: I can't change this. There is no package in apt or yum that will satisfy this requirement. It has to be installed this way. I can make it be a yes/no prompt if you'd prefer or remove it all together. But either way this needs to be done in order for the server to use database functions.

lgsm/functions/install_mta_resources.sh
ChaosMTA: That's not true, if you already have a server and are either moving to a new server, or reinstalling, you will want to move over your resources for your server.

The only reason you should use the default resources is if you are testing. When you want to run an actual server you want to put in your own resources into that folder instead. That's why it asks. If you are setting up a server and intend to put your own resources there, there is no need for LGSM to download them and extract them. It creates extra steps for the user to then remove them and upload their own.

lgsm/functions/info_config.sh
ChaosMTA: The purpose of this is to show more information to the user when they do not have these settings set in their config if they replaced it with their own from the default. So instead of showing "NOT SET", it will actually show what the server is automatically assuming in ./mtaserver details.

I can change this if you wish but I'm just explaining my reasoning.

lgsm/functions/info_glibc.sh
ChaosMTA: This is confirmed working, it used to try and install "UNKNOWN" glibc until i added this commit.

ChaosMTA:
Responded to all your concerns I believe. Let me know if I missed one.

Regarding updates and the script you linked. I can run a command while the server is running called sver to output the server version to console:
```
sver
[04:55:01] MTA:SA Server v1.5.3-release-10614
```

Or automatically on server startup it outputs the current version like so:
```
==================================================================
= Multi Theft Auto: San Andreas v1.5.3 [64 bit]
==================================================================
```

Can you recommend a way I should make an update script to find out the current version?

mogso:
thanks for your request chaos should be handled in 5-10 days

dgibbs64:
@mogso no it will be handled when I get time. Odd how you just got GitHub  today with one comment.

mogso:
@dgibbs64 thanks for your commentt djgibbs64 whoever you are LOL

UltimateByte:
@mogso 

> whoever you are LOL

L-O-L as you said. 
Read the bottom of this page about Daniel Gibbs. https://gameservermanagers.com/

dgibbs64:
#1221

PR #1225, opened 2017-01-03
Fixes Steam Login fail when password includes a '$'

ChaosMTA:
Fixes #1203

PR #1227, opened 2017-01-04
Fixes Steam Login fail when password includes a '$'

### LINKED ISSUE #1203: LGSM Games that require Steam login fail when password includes $
Hello! After trying to install a 7 Day to Die server today, I found out that the reason the install script kept failing was because of a randomly generated password I was using the included a $. 

A simple fix for this problem would be to change the quotes on the variable in the install script from " to '. 

Thanks!

ChaosMTA:
Fixes #1203 

This is branched from develop and merged into develop, versus my other pull request which is a hotfix from the master branch.
https://github.com/GameServerManagers/LinuxGSM/pull/1225


UltimateByte:
Man, you owe me a beer! xD

Issue #1228, opened 2017-01-05
unzip is not a required dependency, yet fn_dl_extract uses it


ChaosMTA:
unzip is used to extract .zip files, yet LGSM does not install or check it as a dependency.

https://github.com/GameServerManagers/LinuxGSM/blob/34d2c96becbe3d7a0c81688ead460cdba17abac9/lgsm/functions/core_dl.sh#L65

dgibbs64:
I dont believe there are any files that use zip within LGSM. I ensured they are all .tar.bz2 because of this zip is not a requirement. Only Unreal used to use zip until I changed it.

ChaosMTA:
But if you have the functionality to extract zip files on that function, you need the unzip dependency.

Zip is required on this pull request:
https://github.com/GameServerManagers/LinuxGSM/pull/1224

dgibbs64:
I see then it would be a requirement for the specific game rather than a general requirement as no servers require it apart from Multi Theft Auto. 

Good work btw. Sorry im quite slow with dev stuff currently I have loads on (still). I will review your work as soon as I actually get some time :)

ChaosMTA:
But then why is tar a LGSM requirement, yet no server on master uses it?

I'm just a bit confused about the inconsistency. I will adjust it to be a MTA requirement only if that's what you'd like though.

UltimateByte:
SteamCMD comes in a tar file if i remember well.
BTW, @dgibbs64 unzip as a requirement is a good idea since it's a pretty common package and mods installations will usually be made from zip files.

cedarlug:
Looks like UT requires it, too:
```bash
cia@cryptbookpro:~/LinuxGSM$ grep -d recurse \\.zip .
./lgsm/functions/install_server_files.sh:               fileurl="http://files.gameservermanagers.com/UnrealTournament/UnrealTournament-Server-XAN-3045522-Linux.zip"; filedir="${tmpdir}"; filename="UnrealTournament-Server-XAN-3045522-Linux.zip";  executecmd="noexecute" run="norun"; force="noforce"; md5="553fed5645a9fc623e92563049bf79f6"
```

dgibbs64:
https://github.com/GameServerManagers/LinuxGSM/pull/1229

PR #1229, opened 2017-01-05
Adds unzip to required dependencies for LGSM

ChaosMTA:
Resolves issue #1228


dgibbs64:
Currently only required by Unreal Tournament and a GTA server release. Adding the requirement to the specific servers would be a good idea as only required by 2 servers currently. I will look at doing this soon

UltimateByte:
Soon, many more will require unzip because of mods-install

dgibbs64:
OK to make things simple I will add unzip to the global dependencies as requested.

PR #1236, opened 2017-01-10
Fixes #1235

### LINKED ISSUE #1235: Find a workaround for Ark config file
The file is located in serverfiles/ShooterGame/Saved/Config/LinuxServer/GameUserSettings.ini
Problem is: the Saved directory doesn't exist until you start the server. So config file cannot be copied to this location.

So either it has to create parent directories first (then config file might be erased upon first start), or, it has to start/stop then replace default config file for LGSM one, then end the install process.

Or maybe you'll find a better solution? Discussion is open.

UltimateByte:
Fixes #1235

Ark had no config file out of the box because its directory didn't exist. Because of that, it wouldn't work out of the box.
Also, LGSM tried to alter config file even if it didn't exist. It's solved by an simple if -f check.

Working on this, I noticed another issue. Let's quote myself:
> Admin password and servername are present into Ark but set within start parameters, so their edition by LGSM is pointless since they will be reset upon server shutdown into the config file, taking start parms values.

So I just removed fn_set_config_vars for Ark.

Ultimately, I have some concerns that need to be checked about Ark's default config file:
Port values might need to be removed from default config file since they are ignored and start parameters are used instead,. On top of that, those are not changed within the config file upon server shutdown to match start parameters, so in the end, if they're not used, they can be misleading. Somebody should test without them into the config file and see if it still works as intended. My guess is it does work perfectly.

### DIFF
--- a/lgsm/functions/install_config.sh
+++ b/lgsm/functions/install_config.sh
@@ -23,17 +23,22 @@ fn_fetch_default_config(){
 # SERVERNAME to LinuxGSM
 # PASSWORD to random password
 fn_set_config_vars(){
-	random=$(strings /dev/urandom | grep -o '[[:alnum:]]' | head -n 8 | tr -d '\n'; echo)
-	servername="LinuxGSM"
-	rconpass="admin$random"
-	echo "changing hostname."
-	fn_script_log_info "changing hostname."
-	sleep 1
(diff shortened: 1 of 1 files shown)


UltimateByte:
OK, relevant remarks, I'll remind them and make another PR.

PR #1241, opened 2017-01-12
Urgent fix for #1239 (broken Steam login)

### LINKED ISSUE #1239: Validate function is not able to login to steam due to bad password
In /lgsm/functions/command_validate.sh

Lines 27 and 29

'${steampass}' needs to be changed to "${steampass}"

Otherwise you won't login to steamcmd due to bad password.

UltimateByte:
Fixes #1239
I just put the single quotes into the right location: The variable. Not the output of the variable, which just prevents it from expanding.

This is an emergency fix since it prevents all game servers that require login from working.

### DIFF
--- a/7DaysToDie/sdtdserver
+++ b/7DaysToDie/sdtdserver
@@ -24,7 +24,7 @@ version="170110"
 
 ## SteamCMD Login | https://github.com/GameServerManagers/LinuxGSM/wiki/SteamCMD#steamcmd-login
 steamuser="username"
-steampass="password"
+steampass='password'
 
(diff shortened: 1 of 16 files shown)


dgibbs64:
Does it fix all issues such as https://github.com/GameServerManagers/LinuxGSM/issues/1203 and the current issue

UltimateByte:
I think so.
#1203 is due to the steampass variable being set in double quotes rather than in single quotes.

Demo:

````
ultimatebyte@game:~/toast$ steampass="$dontlikedollars"; echo "${steampass}"

ultimatebyte@game:~/toast$ steampass='$likesdollars'; echo "${steampass}"
$likesdollars
````

PR #1246, opened 2017-01-14
Fixes Arma3 ports info

### LINKED ISSUE #1240: What's up with Arma3 ports?
Since only one is required and is set with start parms, then why are they into the cfg file?
And especially, why `ERROR! Missing/commented ports in arma3-server.server.cfg` error message even exists, since this is absolutely normal for them to be commented?

First time I'm looking into it, and I feel confused, maybe I'm missing something.
So does this guy with this error message http://steamcommunity.com/groups/linuxgsm/discussions/0/144513128846028225/

https://github.com/GameServerManagers/Game-Server-Configs/blob/master/Arma3/server.cfg#L14
https://github.com/GameServerManagers/LinuxGSM/blob/master/Arma3/arma3server#L31
https://github.com/GameServerManagers/LinuxGSM/blob/master/lgsm/functions/command_details.sh#L521

At first glanc,e I guess port info should come from the main port and some additions set here
https://github.com/GameServerManagers/LinuxGSM/blob/master/lgsm/functions/info_config.sh#L335

UltimateByte:
Fixes #1240 

Those needs to be removed from cfg upon merge to master, that's why I made a hotfix rather than a feature.
https://github.com/GameServerManagers/Game-Server-Configs/blob/master/Arma3/server.cfg#L14-L30

### DIFF
--- a/lgsm/functions/command_details.sh
+++ b/lgsm/functions/command_details.sh
@@ -310,7 +310,7 @@ fn_details_ports(){
 
 	parmslocation="${red}UNKNOWN${default}"
 	# engines/games that require editing in the config file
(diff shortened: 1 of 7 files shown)


UltimateByte:
FUCK!
Committed to wrong branch for my work on mods. -____-

Issue #1186, opened 2016-11-13
Travis Tests failing


dgibbs64:
Looking in to the issue of why travis is failing. I believe it might be due to a recent update to Just Cause 2 server at a guess

	Not Found
	=================================
	/home/travis/build/GameServerManagers/LinuxGSM/tests/serverfiles/Jcmp-Server
		liblua.so => not found
		libsteam_api.so => not found
	/home/travis/build/GameServerManagers/LinuxGSM/tests/serverfiles/libvstdlib_s.so
		libtier0_s.so => not found

I believe the server is not starting because of a dependency issue

here is a working setup

		ldd
		=================================
        linux-gate.so.1 =>  (0xf77c1000)
        liblua.so => ./liblua.so (0xf778b000)
        librt.so.1 => /lib/i386-linux-gnu/librt.so.1 (0xf777b000)
        libsteam_api.so => ./libsteam_api.so (0xf776a000)
        libpthread.so.0 => /lib/i386-linux-gnu/libpthread.so.0 (0xf774d000)
        libdl.so.2 => /lib/i386-linux-gnu/libdl.so.2 (0xf7748000)
        libm.so.6 => /lib/i386-linux-gnu/libm.so.6 (0xf76f3000)
        libc.so.6 => /lib/i386-linux-gnu/libc.so.6 (0xf753d000)
        /lib/ld-linux.so.2 (0x565a8000)
        libstdc++.so.6 => /usr/lib/i386-linux-gnu/libstdc++.so.6 (0xf73c5000)
        libgcc_s.so.1 => /lib/i386-linux-gnu/libgcc_s.so.1 (0xf73a8000)

dgibbs64:
I can confirm that re running tests on passing builds are now failing. I am unsure if there is anything I can do to fix this currently

UltimateByte:
Travis output:

```
Command: ./jc2server start
    fetching check_status.sh...OK
Required status: OFFLINE
Current status:  OFFLINE
New status:  OFFLINE
Test starting:
[ .... ] Starting jc2-server:     fetching check_logs.sh...OK
    fetching check_ip.sh...OK
    fetching check_config.sh...OK
    fetching fix_steamcmd.sh...OK
[  OK  ] Fix jc2-server: Applying steamclient.so general fix: Just Cause 2
    fetching info_config.sh...OK
    fetching logs.sh...OK
failed to connect to server: Connection refused
[...]

PR #1249, opened 2017-01-15
Trying out a hypothesis on the travis failures.

cedarlug:


UltimateByte:
Does the travis test machine lacks a package or something?
If it works only with sudo command available, that's the only reason I can think of.

dgibbs64:
Looks like we have a pass. Well done. What was required to get it to work? Was it just sudo?

dgibbs64:
#1186

cedarlug:
Ultimately just the one-line change to the sudo entry in .travis.yml allows it to go through.

In more detail, there is a large difference between the two environments with just this one simple change.

It appears that permitting sudo allows the development environment to push through an apt-get upgrade.

In two of the latter test runs, I dumped the packages on the system for comparison.

[Here's a diff of the two package lists](https://transfer.sh/mjU2t/lgsm-travis).


I was expecting at most a dozen packages to be different between the two development environments, but 457 differences between the packages installed with/without sudo was surprising.

cedarlug:
`sudo: required` == travis success. #1186

UltimateByte:
You must be a living god for f**k sake.
That's a one liner fix, but amazing work behind it!

PR #1252, opened 2017-01-15
Trial for yaml parser

dgibbs64:
yaml could be the future method for managing LGSM config


dgibbs64:
yaml is not what we will be pursuing. But rather using bash variables instead. Yaml did not provide enough benefit for the project.

Issue #631, opened 2016-01-16
Integration with mods for applicable games


KroniK907:
Many server admins want to have an easy way to add/remove SourceMod plugins. Would it be possible to add support for managing SourceMod plugins through LGSM? Or at least having the ability to install SourceMod and MetaMod by default for any games that support it? 

It would be cool to be able to have a package manager for SourceMod plugins that was a part of LGSM that would not only install the plugins but also create the default CVARs etc in the config file.

UltimateByte:
I've been that idea for a while, it could definately happen, yes.
It could also propose, ULX for gmod.
2 things to face btw : 
- Find a reliable lastest version link for them
- Update them without losing settings, even if the settings file changes. (i think about sourcemod and the way it manages admins...)

jaredballou:
There really isn't even a "SourceMod official" installer, so this would be waaaaaay more work than we can reasonably take on. I built something much simpler just for my own plugins - https://github.com/jaredballou/insurgency-sourcemod/blob/master/tools/download-plugin.sh - it finds the list of plugins I created, then downloads the updater files and sees what is needed to install, and pulls from Git. I just threw it together in an hour or two, so it's pretty raw, but if you had a list of known good plugins we might be able to leverage it that way. My guess is that you're going ti have to just download them on your own, however, unless there is an API to the AlliedModders forums that we could use to do this.

KroniK907:
Even without the the package manager, installing the current version of SourceMod and MetaMod by default when installing a compatible game would be a nice step up. The plugins are fairly easy to manage but getting SourceMod installed for the first time can be a bit confusing.

UltimateByte:
We still have to figure out a reliable way to get the latest version at any given time. It's true that sourcemod don't seem to provide an "always up to date link".

jaredballou:
So, Christmas comes early for you guys. I created a new function, install_sourcemod.sh, that does the following:
1. Checks ${systemdir}/addons/metamod and confirms install if existing
2. Downloads MetaMod download page, finds mirror list
3. Tries each mirror until file is downloaded
4. Unzips MetaMod to systemdir
5. Checks if SM is installed same way, same confirmation
6. Finds latest SM version from pointer file, downloads it
7. Unzips SourceMod to systemdir

[...]

PR #1275, opened 2017-01-27
Rust wipe & permissions on /sys

### LINKED ISSUE #1254: RUST & some other games - wrong permissions on /sys = crash on boot
Symptom: the game server just stops while loading after `Initializing 74 stability supports`

Update3: It's actually not a kernel itself issue, but rather an issue because of permissions on `/sys`

Fix thanks @cedarlug :
(with elevated privileges)
````bash
chmod a+rx /sys /sys/class /sys/class/net
````

### LINKED ISSUE #1260: command: ./rustserver wipe
Server wipe means clear game data. This task is usually required after monthly updates. 
Since some patches don't require a wipe, we cannot add an option for automatic wipes upon updates.
Would be a nice addition for Rust server owners to have a command for that.

````bash
./rustserver wipe
````

Update: There is now only one file to wipe.
Example: proceduralmap.3000.736413606.140.sav

So basically, this is gonna be like:
````bash
rustwipemaps="$(find "${serveridentitydir}" -name "proceduralmap.*.sav)"
if [ -n "${rustwipemaps}" ]; then
	echo "These file(s) will be removed:"
	echo "${rustwipemaps}"
	echo "Continue ? y/n"
	rm "${serveridentitydir}/proceduralmap.*.sav"
else
	echo "Nothing to wipe"
fi
````

UltimateByte:
Fixes #1254 permissions on /sys
Sample output with & without sudo access:

sample

Imrpoves #631 (mods integration)

Fixes #1260 Rust server wipe feature

I don't have time at the moment for proper output & logging for the wipe command (command_wipe.sh). @dgibbs64 If you can do this, let's go, otherwise i'll do it during the night, I think.


UltimateByte:
Sample output for wipe command (Rust only, but made it so that other games can be added quite smoothly):

wipe command

PR #1283, opened 2017-01-31
Minecraft fix for improper default executable

ChaosMTA:
I have no idea how this was not noticed before.
Without this change the following will happen:
[ FAIL ] Starting mc-server: Executable java -Xmx1024M -jar minecraft_server.jar was not found

### DIFF
--- a/Minecraft/mcserver
+++ b/Minecraft/mcserver
@@ -86,7 +86,7 @@ filesdir="${rootdir}/serverfiles"
 ## Server Specific Directories
 systemdir="${filesdir}"
 executabledir="${filesdir}"
-executable="java -Xmx${javaram}M -jar minecraft_server.jar"
+executable="java -Xmx${javaram}M -jar ${filesdir}/minecraft_server.jar"
 servercfg="server.properties"
 servercfgdefault="server.properties"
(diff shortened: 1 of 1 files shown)


UltimateByte:
https://github.com/GameServerManagers/LinuxGSM/blob/develop/lgsm/functions/check_executable.sh#L11

`if [ ! -f "${executabledir}/${execname}" ]; then`  
Would expand to something like:  
`"/home/minecraft/serverfiles/java -Xmx1024M -jar /home/minecraft/serverfiles/minecraft_server.jar"`

Would it work with the first part of the expand being totally wrong?

ChaosMTA:
Interesting. I'm not sure why that check doesn't work right now without the filesdir line I added.

UltimateByte:
We should add the full path of the non found executable error message.

ChaosMTA:
What do you mean @UltimateByte

UltimateByte:
Well, the non found executable error message that i made just shows the not found file's name. It should tell the full path. I'm still concerced about my first message though, i'll have a look if needed  Sleeping a bit for now :)

UltimateByte:
Fixed #1285

PR #1304, opened 2017-02-06
Edit defaultmap for CS:GO active duty map change

stubbegianni:
Dust2 has been taken out of active duty. Changed map to mirage and changed mappool to mg_active following the Valve documentation: 
https://developer.valvesoftware.com/wiki/Counter-Strike:_Global_Offensive_Dedicated_Servers#Starting_the_Server


UltimateByte:
Thought Dust 2 had been removed entirely, too bad :p 
Thumbs up for Mirage as a default map though.

stubbegianni:
I did not know what was the best option to pick as default map, but it seems mirage is pretty popular, other option would have been cache

UltimateByte:
Any map that isn't Dust is fine to me 👍 
If I had to chose, would be Inferno or Nuke :p

PR #1310, opened 2017-02-09
Feature/tuserver

dgibbs64:


marvinl97:
As far as I know the GSLT for Tower Unite can't be set as start parameter. It has to be set in the config: https://github.com/GameServerManagers/Game-Server-Configs/blob/master/TowerUnite/TowerServer.ini#L4

PR #1326, opened 2017-02-19
Fix logs keep being created with some games with a more general way of handling logs

UltimateByte:
`gamelogdir `needs to be created if variable is set, otherwise `check_logs.sh` will attempt to create it upon almost any command if the server wasn't started yet and directory doesn't exist, as per https://github.com/GameServerManagers/LinuxGSM/blob/master/lgsm/functions/check_logs.sh#L24

symlink is done if `gamelogdir` variable is set rather than checking for what game it is, and after gamelogdir has been created (this order doesn't change anything, just a bit more elegant).

To sum up: Working with `if [ -n $"{gamelogdir}" }]` is a more general and simple way of doing it and shouldn't cause any kind of issue as long as gamelogdir variable is set to the proper path.

Thanks @Khelgar for reporting the issue

### DIFF
--- a/lgsm/functions/command_console.sh
+++ b/lgsm/functions/command_console.sh
@@ -16,7 +16,6 @@ echo ""
 if ! fn_prompt_yn "Continue?" Y; then
 	echo Exiting; return
 fi
-done
 fn_print_dots "Accessing console"
 sleep 1
 check_status.sh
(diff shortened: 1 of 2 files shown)


UltimateByte:
Oh, console is broken because there is a "done" that shouldn't be here.
Fixing it as well.

Issue #1324, opened 2017-02-19
Sort out core_getopt.sh


dgibbs64:
Get opt is getting a little messy becuase of all teh different variants. This needs a tidy up od the code

UltimateByte:
First :p 
https://github.com/GameServerManagers/LinuxGSM/issues/1261

UltimateByte:
I'm on it.

Issue #1322, opened 2017-02-19
Add option to remove non-compressed files when using FastDL


NickParks:
While it doesn't matter much, it would be a nice option that if the user is using bzip2 compression when generating FastDL files, to remove the original file that is not compressed.

Not too sure if GMOD/Source will download both files or take only the compressed one (haven't tested it myself tbh). But either way it can't hurt to have the option :)

~ Nick

UltimateByte:
I think I had a good reason for not removing them, because I remember myself having the same idea as you. I would have to dig back into it to see if it was really a good reason or not. :o))

UltimateByte:
![option fastdl gmod](http://i.imgur.com/stc2ZA6.jpg)

UltimateByte:
Will be added upon next merge

UltimateByte:
Closing to clean my tasks list.
Code is ready to be merged into develop branch for next release.

UltimateByte:
Just a follow up: FastDL will change a lot. And actually, uncompressed files will be removed automatically since they are not necessary.

Issue #1261, opened 2017-01-18
core_getopt.sh needs rework to avoid duplicates


UltimateByte:
Loads of lines in core_getopt.sh are just duplicates.
It's kind of a mess to manage.

I'll try something with arrays.

UltimateByte:
Well, i transformed 985 code lines into 174, while gaining a lot of simplification, IMO.
https://github.com/GameServerManagers/LinuxGSM/blob/feature/ultimatebyte-0220/lgsm/functions/core_getopt.sh

This will need testing with all games that have special commands to ensure it is working properly. Should be pretty fine though.

I also changed this unreadable blue color to cyan, as per https://github.com/GameServerManagers/LinuxGSM/issues/1169 since now it's just one single variable to edit in order to achieve that ! :p

UltimateByte:
Closing to clean my tasks list.
Code is ready to be merged into develop branch for next release.
This would need to be re-opened if my code isn't accepted.

Issue #721, opened 2016-02-26
Extend FastDL to other source games


UltimateByte:
fastdl command is released for gmod (closed the old issue)

Now, need to adapt it for other source games.

UltimateByte:
![screenshot](http://i.imgur.com/2Z6VRHd.jpg)

UltimateByte:
Closing to clean my tasks list.
Code is ready to be merged into develop branch for next release.

PR #1331, opened 2017-02-22
New core_getop.sh and FastDL generator for all source servers

UltimateByte:
- New core_getopt.sh with no redundant content, as per #1324 and  #1261
- FastDL extended to other source games as per #721
- Default remove non-compressed FastDL files #1322
- Fixed BladeSymphony missing GitHub Info (thanks @cedarlug )

Notes: Please double check and try the new opt on multiple games before final release. We need to make sure that no commands are missing or added while unavailable, even though i spend a lot of time working on it to match previous opt.
This new opt uses a bit more cpu time because of loops and arrays. It should be fine even on slow machines but it's worth testing if you got one available, and it's worth seing if you got ideas to optimize it a bit more, which is why i added 4 reviewers.

Cheers 🍻


UltimateByte:
**Fixed**
- I noticed dev commands didn't have a core_exit.sh, so I added it.
- core_exit.sh wasn't placed at the right location in core_getopt.sh (my bad), which made the "uf" command display an unnecessary opt after a successful functions update or dev commands.
- Lua enforcer should touch the file before echoing into it.
- Lua enforcer had to look look for non .bz2 files, not the opposite.

**Remaining issues**
- "addons" dir isn't scanned: this is problematic, especially for Garry's Mod.
- For the explanation right below, the directory structure you made will sadly not work for Garry's Mod.

**About Garry's Mod file structure:**

For Garry's Mod, FastDL is more about GUI textures, models, effects, sounds, than maps. Those are either coming from the Workshop (in which case, FastDL won't help), or from addons extracted into `garrysmod/addons/*`
All addons in `addons/<addon_name>` are virtually mounted to `garrysmod` directory, which means that the game acts as if they were extracted to the game directory. The same goes from workshop's .gma files. But they are not actually extracted, and there is a good reason for that: allow to install many addons while keeping all of their files separated, and allow easy addon removal.

Garry's Mod addons can contain any file type, organized in any way that is up to the addon's author. 
It will always contain a `lua` dir for addon's script files, then other dirs might or not use standardized names. Usually, we will find `materials`, `models`, `particles`, `sound`, etc. anyways, which themselves will contain sub-directories, that are sometimes the addon's name (to prevent taking the lead over original game files in the case of duplicate file names) or some more generic names. 

In any case, for the addon to be able to use these files from the FastDL, the original directory structure must be kept. It is unfortunately not as simple as "`*.ext` files in `whatever` dir go to `fastdl/whatever` dir". 

To sum it up: Consider that any file type can be found in any directory, and especially in `addons`. 
So the only way, or at least the easiest to work with Garry's Mod, is to do what I did before: 
1) Scan the whole game directory and copy any file that matches extensions into the "fastdl" dir while keeping the paths structure.
2) Move `fastdl/addons/*/*` to `fastdl`. 
3) Move any remaining darkrpmodification addon stuff in `fastd/lua/"*` to `fastdl`

**Remaining question about Garry's Mod's file structure**

- Is it sometimes useful to keep `fastdl/addons/<addon_name>/*` or should those subdirs always be moved to the `fastdl` dir directly? If they should always be moved, then we can avoid the prompt about it.

**Final thoughts**

As I said once (should have insisted more on that) Garry's Mod is totally different and should have its own FastDL dir build process. Hope that this time, I provided enough details!
I need some sleep atm, but when I wake up, I'll see what I can do to try making a code that achieves this while still using your great additions. Of course, if you wish to work on this before I wake up to make sure we're ready for the release, don't wait for me!

Cheers 🍻

UltimateByte:
`webdir="${lgsmdir}/public_html"` means it won't work with default userdir apache2 mod if user installed his game directly in the user's home dir as suggested.

dgibbs64:
@UltimateByte thanks I didnt know that

UltimateByte:
Your last commit was amazing, it works as it should!

What I've tried so far:
Running a gmod server with ACF addon, built the FastDL, then
- Tried with ACF in my client: Didn't donwload much files
- Tried without ACF in my client: Downloaded all required ACF content.
- Removed the server's lua enforcer, cleaned my client's downloads and tried again: No more FastDL downloads happened.

Works as expected.

Suggestion: 
- Provide less user output while removing or compressing files.

Concern: 
- Maybe files in `addons` dir should be scanned for other Source games to have their addons content downloaded using FastDL.

UltimateByte:
It's all good to me now, except for my little upper concern. :)

dgibbs64:
I can add the addons dir as well for other source engine games. There is no harm in having them there anyway I dont think

PR #1348, opened 2017-03-02
[critical] Fixing ts3db prompt's opposite behavior

UltimateByte:
Fix was super simple, but issue is super annoying for ts3server users.


UltimateByte:
I'm merging this very small fix to master. Hopefully i won't break the release pattern or something.

PR #1397, opened 2017-04-10
Add tmux dimensions

### LINKED ISSUE #1396: [FEATURE REQUEST] setting the tmux dimension at launch
currently LGSM has no way to specify the dimensions of new tmux sessions, and as such spawns detached sessions @ 80x23.

at the moment i have only found 2 ways to remedy this, either modifying the command_start.sh file to include -x and -y, this works but is overwritted when when LGSM updates this function.

the other method i found is to create a simple bash script to act as a shim for the tmux command, this works but with some games, permission errors can arise.

im not sure on the exact reason behind the permission problem, however i have only noticed it with arma3 so far, tf2 seems unphased by it.

so long story short, can we have the ability to set the detached sessions width and height at session start, either by a variable in the lgsm script, or even a system wide variable.

TechNZ:
added the ability to set the default width and height of the detatched session using variables tmux_width and tmux_height.

these variables may be set at a local level or system wide.

the script checks if the input from the variables is a valid positive integer and if input is not a valid positive integer, sets the values for the tmux to the default values of 80x23

Fixes #1396

### DIFF
--- a/lgsm/functions/command_start.sh
+++ b/lgsm/functions/command_start.sh
@@ -56,6 +56,20 @@ fn_start_teamspeak3(){
 
 fn_start_tmux(){
 	fn_parms
+	
+	# check for tmux size variables
+	if [[ ${tmux_width} =~ ^[0-9]+$ ]]
+	then
+		tmux-x=${tmux_width}
+	else
+		tmux-x=80
+	fi
+	if [[ ${tmux_height} =~ ^[0-9]+$ ]]
+	then
+		tmux-y=${tmux_height}
+	else
+		tmux-y=23
+	fi
 
 	# Log rotation
 	check_status.sh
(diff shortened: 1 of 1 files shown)


UltimateByte:
Much better :p 
I suppose @dgibbs64 will accept this, since it doesn't change anything for other users.
We probably should add these parameters to all gameserver files though.

Forgot about branching though.
Basically, this PR should be made to "develop" branch and preferably named "feature/tmux-123" (with 123 being any set of numbers). More info here: https://github.com/GameServerManagers/LinuxGSM/wiki/Branching

dgibbs64:
Looks good to me :). This will be in the next release.

TechNZ:
Many thanks for the assistance @UltimateByte , this is my first time using github for any form of developement purpose. so it was a good learning experience.

And many thanks @dgibbs64  for accepting.

Issue #1372, opened 2017-03-16
[Dependency] "bc" package


UltimateByte:
`bc` package is needed for some commands to work properly. This one isn't installed on minimal distros.
Would be nice to start taking take care of required packages that might not be default installed on minimal distros, since people using them don't always know how to figure out missing packages.
![pic](http://i.imgur.com/0B6oQft.png)

dgibbs64:
we can get this added in.

UltimateByte:
Another one to add: `binutils`
Required for `strings` command. used between other things into `install_config.sh`

dgibbs64:
All added

PR #1414, opened 2017-04-28
Update check_deps.sh

lanrenags:
binutils not part of the standard installation of Ubuntu (tested on 17.04)

"strings"-command is missing and used in install_config.sh


lanrenags:
There is the same rewquest on this ticket

[Dependency] "bc" package #1372 maybe also add 'bc'  to "array_deps_required"

Issue #1418, opened 2017-04-29
Add support for factorio experimental builds 


Duckbuster:
Add support for factorio experimental builds

skynett81:
1.stop your server . 
2. https://www.factorio.com/download-headless/experimental
3. copy the files to serverfiles and overwrite . 
4. start your server and you have 0.15.4

Duckbuster:
Sure, but they're currently releasing almost a patch everyday. So it becomes a bit cumbersome.

UltimateByte:
https://github.com/GameServerManagers/LinuxGSM/pull/1422

PR #1422, opened 2017-05-02
#1418 factorio experimental

Digital-Storm:
Added Factorio Experimental build support. Needed to update the core_dl to extract the file properly, as well as added a new variable to switch from "stable" to "experimental". Updated version checking to return 1 line instead of more than 1, which would break the comparison as well.

### DIFF
--- a/Factorio/fctrserver
+++ b/Factorio/fctrserver
@@ -49,6 +49,7 @@ channeltag=""
 
 ## Updating | https://github.com/GameServerManagers/LinuxGSM/wiki/Update
 updateonstart="off"
+updatebranch="stable"
 
 ## Backup | https://github.com/GameServerManagers/LinuxGSM/wiki/Backup
 maxbackups="4"
--- a/lgsm/functions/core_dl.sh
+++ b/lgsm/functions/core_dl.sh
@@ -60,6 +60,8 @@ fn_dl_extract(){
 		tarcmd=$(tar -zxf "${filedir}/${filename}" -C "${extractdir}")
 	elif [ "${mime}" == "application/x-bzip2" ]; then
 		tarcmd=$(tar -jxf "${filedir}/${filename}" -C "${extractdir}")
+	elif [ "${mime}" == "application/x-xz" ]; then
+		tarcmd=$(tar -xf "${filedir}/${filename}" -C "${extractdir}")
 	elif [ "${mime}" == "application/zip" ]; then
 		tarcmd=$(unzip -d "${extractdir}" "${filedir}/${filename}")
 	fi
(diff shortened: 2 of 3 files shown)


dgibbs64:
Very happy with this well done!!. merging in to develop branch ready for next release

ClaytonSmith:
Is there documentation on how to use this new feature?

wizpig64:
@ClaytonSmith Here's what worked for me:

1. Download the fctrserver script [from the develop branch](https://github.com/GameServerManagers/LinuxGSM/raw/develop/Factorio/fctrserver) rather than from master.

2. Change the line `branch="stable"` to `branch="experimental"`.

3. Change the line `githubbranch="master"` to `githubbranch="develop"`. Once this gets released into master you can switch it back if you want.

4. Configure and run everything else [as you would normally](gameservermanagers.com/lgsm/fctrserver/).

Digital-Storm:
@ClaytonSmith As @wizpig64 said, though do note, that gibb made a change to the variable from updatebranch to branch, so you may need to modify the lgsm/functions/factorio_update.sh script, and replace "updatebranch" with "branch". Do a search in that file for "updatebranch" and change it to "branch" and you should be all set until he merges my patch.

wizpig64:
good catch @Digital-Storm, thanks.

ClaytonSmith:
Good to know, thanks. 

Just a side note. I feel it would be a bit cleaner to be able to specifying the target branch or a specific release number through command line argument, `./fctrserver update x.x.x` or `./fctrserver update experimental`.

Maygus:
@wizpig64 the dev branch link isn't working?

Maygus:
I found - looks like you're changing how the installer first downloads. However...

I cannot find the `branch="stable"` entry you mention... nor any reference to `updatebranch="xxxx"` will this still work?

Digital-Storm:
@Maygus You will need to download and overwrite your original ./fctrserver script with the development branch version, as doing a ./fctrserver update with the development branch selected, will not update the ./fctrserver script itself. 

Perhaps that is something we can get included in the update script.

Maygus:
Thanks - yeah I figured it out after digging around. Had to download lgsm.sh and run that which pulls down the factorio script (when selected) and then digging in it found where I add the entra for experimental so that the factorio script can look at latest experimental. :)

XGhozt00:
Does this still work? I have:

`branch="experimental"`

Inside: `./lgsm/config-lgsm/fctrserver/fctrserver.cfg`

It's using the port and rconpassword variables correctly from that file, but I can't get `./fctrserver update` to install the latest experimental version?

PR #1449, opened 2017-05-27
Add squadserver

Scarsz:
Requesting that @Odinthewanderer also give this a try if possible


dgibbs64:
I will merge in to develop then merge across to config and I will use squad as a test for the new release

PR #1508, opened 2017-06-29
Changed tar exclude to be the configured dir

Drizzt321:
Updated the tar exclude to be the configured backup dir, instead of just a hardcoded "backup". This allows for people to change the backup directory location to still be in the game server dir, but in a custom directory. For me specifically it's an NFS mounted directory to backup to my NAS with the cloud backup then enabled on that dir.


Drizzt321:
Strange. If I'm reading it right the test scripts were having trouble simply fetching https://raw.githubusercontent.com/GameServerManagers/LinuxGSM/master/lgsm/functions/core_dl.sh? So it's not a problem with my PR?

UltimateByte:
Don't worry about Travis test, it's fixed in the dev branch :)

Drizzt321:
Ha! Ok :)  I'm also going to do some comparisons with bzip2 vs gzip, and various compression options to see about time and output file size if there's much of a difference. Not in this PR though.

UltimateByte:
Excellent idea :)

Issue #1517, opened 2017-07-04
DOI Setup Broken


naxIO:
Im trying to install a Day of Infamy Server on Ubuntu 16.04.1 LTS and I keep getting this. I followed the required setup.

```
=================================
Day of Infamy
Linux Game Server Manager
by Daniel Gibbs
Contributors: http://goo.gl/qLmitD
https://gameservermanagers.com
=================================

Server Directory
=================================
Warning! A server is already installed here.
/home/doiserver

Continue? [Y/n] Y

Creating log directories
=================================
installing log dir: /home/doiserver/log...OK
installing LinuxGSM log dir: ...mkdir: cannot create directory '': No such file or directory
FAIL
```

SkullMasher:
Not sure if related but I got an error updating DOI  earlier this week. Solved it by doing `./doiserver uf` several times then `./doiserver r`. Hope that helps.

UltimateByte:
I know this bug, I solved it for some other games already.
It's just a matter of directories variables to fix within LGSM code.
I'll have a look when i got time if nobody fixes it before me.

naxIO:
I see thank you! Maybe it would be a good idea to separate the core logic of LGSM from actual game libraries?

bromfj:
I will submit a fix for this soon. But it can be fixed by adding the following variable to 'doiserver':

lgsmlogdir="${logdir}/lgsm"

jaredballou:
Really shouldn't lgsmlogdir be defined in linuxgsm.sh for all games?

bromfj:
Yes it should and the fix is submitted for linuxgsm.sh, but this will resolve the problem for people who have already tried to install it.

naxIO:
thanks!

PR #1520, opened 2017-07-08
Fix for issue 1517 'DOI Setup Broken'

bromfj:
When attempting to install Day of Infamy it will fail with the following:

Creating log directories
=================================
installing log dir: /home/doiserver/log...OK
installing LinuxGSM log dir: ...mkdir: cannot create directory '': No such file or directory
FAIL

---

The 'lgsmlogdir' variable is missing in doiserver.


dgibbs64:
#1517 fixed

Issue #1524, opened 2017-07-12
[BUG] Notifications via PushBullet post twice in one message.


Shigbeard:
When the script attempts to post a notification to PushBullet, for instance in the event that the server crashes, it will post it's message twice in one line. See the below image.

![http://img.ndlee.com/shigbeard/onhigj5.png](http://img.ndlee.com/shigbeard/onhigj5.png)

This issue is caused in alert_pushbullet.sh, by having fn_rawurlencode() not only echo the resulting string, but also return it. I was able to eliminate the issue by editing the file and removing Line 27 (`REPLY="${encoded}"`) and editing lines 30 and 31 to read...

```bash
pbalertbody=${fn_rawurlencode "${alertbody}")
pbalertsubject=$(fn_rawurlencode "${alertsubject}")
```

This may not be the correct way of fixing it however, and someone may have a better solution.

Shigbeard:
Pull Request merged, issue resolved. Closing issue.

dgibbs64:
Will be on the next release

PR #1525, opened 2017-07-12
Fixed PushBullet alerts duping strings, see #1524.

Shigbeard:
As outlined in GameServerManagers/LinuxGSM#1524, the PushBullet alert script was somehow printing the same text twice in a single field. By removing the reply variable from both the fn_rawurlencode() function and from the variables used not 3 lines down, we produce the intended alert.

I made a minor error in a comment where I mislabeled echo as being faster, however the script functions as intended.


dgibbs64:
Thanks for this. Will be merged when I can!

PR #1534, opened 2017-07-16
Adding Telegram Bot Alerts

bennet0496:
I added a feature to send alerts via a telegram bot. Alerts can be sent to private chats and groups.

To use this feature the user would have to create a telegram bot as discribed [here](https://core.telegram.org/bots#6-botfather). The api key has to be set as `telegramapikey`. The `telegramchatid` is the uniq identifier for the target chat which is either the integer chat id or the name of a channel (in `@channelname` format).
To get the chat id you would need to send a message to your bot and check your bots new messages with `curl -X POST https://api.telegram.org/botXXX:YYYY/getUpdates` (where XXX:YYYY is the api key). This will an json string which contains the chat id (e.g. `..."chat":{"id":-202081584,"title"...` the chat id would be `-202081584`).

A message would look like this 
image


dgibbs64:
This is awesome thanks!!

PR #1539, opened 2017-07-20
Symlinking the SteamCMD dir for ARK

Drizzt321:
Symlinking the SteamCMD directory into the correct ARK directory so that the mods auto-management will work fine. See http://ark.gamepedia.com/244.3 for the CLI option and Game.ini configuration settings.


dgibbs64:
Thanks for this. I think this will have to come under fix_ark.sh as this will need to roll out to every install and not just new ones

Issue #1538, opened 2017-07-19
[DST] Master+Caves


bugavka:
i want 1 serverfiles 2 instance. when i do 2 service dstserver and dstserver-2, they both apply setting from /lgsm/config-lgsm/dstserver/dst-server-Master.cfg.
i can make setting by hand, but on start, for example:
dstserver, if in dst-server-Master.cfg shard="Master", run shard Master - ok
dstserver-2, in dst-server-Master.cfg all the same shard="Master" of course runs shard Master - not ok.
dstserver-2 ignoring any cfg besides dst-server-Master.cfg.
if i before start dstserver make dst-server-Master.cfg how i want, then start dstserver, then change dst-server-Master.cfg for dstserver-2, run dstserver-2. Servers works ok! How is do it automatic?

UltimateByte:
Just duplicate your config file with the right name.
Doc:
https://github.com/GameServerManagers/LinuxGSM/wiki/Multiple-Game-Servers
https://github.com/GameServerManagers/LinuxGSM/wiki/LinuxGSM-Config-Files

bugavka:
in csgoserver then i do second time - ./linuxgsm.sh i, script do csgoserver-2 and csgoserver-2.cfg and all works fine. In dstserver - first time - dstserver and dst-server-Master.cfg, in second time do nothing, I try diplicate config file, what name is right? i try:
for dstserver-2
dstserver-2.cfg, dst-server-dstserver-2.cfg, dst-server-2.cfg. dstserver.sh this files ignore.
if qustion only in name, what config file name is for dstserver-2? Please.
and thanks for help, very much.

Fedcomp:
I have same issue. Can't find a way to create proper config for it. What is the reason there is special behavior which doesn't match wiki page?

marvinlehmann:
@Fedcomp #1554

Fedcomp:
This is the first result in google and so i will post temporary solution:
open `lgsm/config-default/config-lgsm/yourserver/_default.cfg` and comment out
```sh
servicename="dst-server-${shard}"
```

marvinlehmann:
Because there the config structure was completely different when this was implemented.

adrensnyder:
Thanks @Fedcomp for you suggestion. I make the Cave work.
Hope that will be fixed soon :-)

UltimateByte:
Needs merge. :)

dgibbs64:
[...]

PR #1554, opened 2017-07-29
Added pre-install config creation

marvinlehmann:
Partly adressing the feedback by @Khelgar1:

> The instructions for installing a 7dtd server at https://gameservermanagers.com/lgsm/sdtdserver/ are incorrect. You cannot edit the common.cfg file because the lgsm directory structure isn't created until after you run ./sdtdserver install command once... Am I missing something in how this should be done because it's not at all obvious.
> IMO what should happen when you run ./linuxgsm install is that is should create the proper directory structure and then then download the sdtdserver script. Downloading it before the folders are created is just confusing.
> Finally the message Installed 7 Days to Die server as sdtdserver is very misleading it would be clearer if it said Downloaded sdtdserver script then give the message To install server run ./sdtdserver install or something along those lines
> _Finally Finally it might be useful to add an instruction in how to create an entire common.cfg file for the first time cat \_default.cfg >> common.cfg works a treat for me. Then it's easy to edit common.cfg to suit. I've not even looked through common.cfg apart from entering the steam details though so maybe you have reasons for not doing this, apart from overwriting existing cfg files_

Furthermore it'll fix and break the current already broken DST sharding stuff (#1538).

This is **NOT** completely tested yet and I'd like to get some feedback if it's OK.


lalmeras:
I can confirm the issue caused by servicename= setting in _default.cfg. As _default.cfg is one of the first file sourced in dstserver, it forces the instance config file to be dst-server-Master.cfg (as at the time _default.cfg is sourced, $shard is Master).

Not to set servicename as proposed allows correct configuration loading.

And for the moment, as ./dstserver install automatically triggers _default.cfg refresh, it is not even possible to fix it in lgsm/config-default as it is systematically overwritten.

I'll try your patch and let you know if it allows me to configure sharding.

lalmeras:
On my side, with the branch config-dl:

* `./linuxgsm.sh` dstserver creates now both dstserver script (then dstserver-2, ...) and a default config in lgsm/config-lgsm/dstserver (_default.cfg, common.cfg, dstserver.cfg)
* but for multi-instance, `./dstserver(-2) install` is correct only if `servercfgdir` and `servercfgfullpath` are redefined each time in `<instance>.cfg`; this is because these variables references directly or indirectly ${shard}, that is `Master` at the time _default.cfg is read. So these variables must be redefined in `<instance>.cfg`
* with these redefinition, install phase generates expected `.klei/DoNotStarveTogether/Cluster_1` configurations
* a fix may be to use functions instead of variables for `servercfgdir` and `servercfgfullpath` (as it is currently done for `parms` and `fn_parms`)
* currently, this patch break existing installations, as they rely on an out-of-spec `dst-server-Master` file. I can't find any solution except to add this ugly workaround at the end of `_default.cfg`
```
if [ -f "${configdirserver}/dst-server-Master" ]; then
  servicename="dst-server-${shard}"
fi
```

My files:

dstserver.cfg
```
sharding="true"
master="true"
shard="Master"
cluster="Cluster_1"
cave="false"

servercfgdir="${clustercfgdir}/${shard}"
servercfgfullpath="${servercfgdir}/${servercfg}"
```

dstserver-2.cfg
```
sharding="true"
master="false"
shard="Slave"
cluster="Cluster_1"
cave="true"

servercfgdir="${clustercfgdir}/${shard}"
servercfgfullpath="${servercfgdir}/${servercfg}"
```

Hope that helps

marvinlehmann:
Exactly my experience. Thank you for confirming and writing everything detailed down! =)

But I can't think of nice solution too :/

miztertea:
for 7dtd server page on website, could you not just add this command to the instruction right before the nano one?

mkdir -p lgsm/config-lgsm/$gameName/

Sdrabor:
Just to note... this did cause a slight hiccup in the setting up of a new server for me tonight.  Wasn't too much of an issue after figuring out I could run the install and then just run it again a second time after adding the steamCMD login details to common.cfg.

marvinlehmann:
I think we could merge it but it'll break existing DST servers.. which probably aren't working properly anyway. @dgibbs64 do you have download stats for DST?

When it gets merged, I'll edit the wiki to point out the redefinition/"override" of the servercfg paths in the instance.cfg.

dgibbs64:
@marvinlehmann would you be able to re-review this once you are happy I can merge this

marvinlehmann:
@dgibbs64 I'll take a look at it and will adjust it to the current version. :)

marvinlehmann:
@dgibbs64 It should be at the same functional state as before now.
Not perfect because of the issues mentioned in this conversation. :)

dgibbs64:
Thanks. I am taking a look. Please could you add some comments to make it easier for me to understand what the code is doing?

marvinlehmann:
Almost all of the code was originally there.
e.g. fn_install_default_configs() basically consists of https://github.com/GameServerManagers/LinuxGSM/blob/develop/linuxgsm.sh#L316-L366

To fn_install_file() I just added a call for fn_install_default_configs(), some messages and I renamed "local_filename" to "serverscriptname" because it would be overridden by the fetch functions, I believe.

That's basically it. Nothing fancy. Just reusing existing stuff. =)

PR #1595, opened 2017-08-28
Backup exclusion testing

jimwald1221:
I've reworked the backup command to use a flat file that is used for file and folder exclusions in an attempt to allow admins to exclude files or folders from their backups. The idea being they can exclude game files that will be replaced by repairing the game server and it helps keep their backups smaller.

I've also added the necessary exclude file variable and a check for the "backupexclude.cfg" in the same fashion as "common.cfg" and "instance.cfg" so it should fold into current installations without issue.


dgibbs64:
I will have to take a look at this when I get a chance 👍

jimwald1221:
Yeah no rush on it. It has been working well for what I'm using it for and wasn't sure how much other admins would like such a feature but figured I push it to you to see what you thought.

jimmyatSplunk:
Interested in the status of this. Been migrating my backup strategy to push the archives to an S3 bucket and wanted to see about this exclusion method so I can update my LGSM configs to the newest version without worrying about `command_backup.sh`

jimmyatSplunk:
Actually looking at the age and the changes of this request, it may be best to redo the request with the more up-to-date version of the app.

Frisasky:
close this due to we have a much better and up-to-date implement #2312

PR #1614, opened 2017-09-13
Fix simple bug [gmod]

Doc1337:
DarkRP is gamemode, not addon.


dgibbs64:
Thanks this will be in the next release

PR #1652, opened 2017-10-15
changed all shebangs to use /usr/bin/env to improve compatibility

The-M1k3y:
By changing the shebangs the scripts can be used on systems with non-standard paths without the need to modify them.

### DIFF
--- a/lgsm/functions/alert.sh
+++ b/lgsm/functions/alert.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # LinuxGSM alert.sh function
 # Author: Daniel Gibbs
 # Website: https://gameservermanagers.com
--- a/lgsm/functions/alert_discord.sh
+++ b/lgsm/functions/alert_discord.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # LinuxGSM alert_discord.sh function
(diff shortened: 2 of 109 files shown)


dgibbs64:
I am unsure if this is the best option
https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my

UltimateByte:
I can't think of a reason why bash would be installed in a non-standard path. Could you provide an example?

The-M1k3y:
I'm currently setting up my servers with NixOS, which has the bash executable at /run/current-system/sw/bin/bash.
I understand if you don't want to use the env-shebang, but as far as I can see there are no real disadvantages when using env.

UltimateByte:
Well, that is something to test on supported distro First, including legacy distro. If that only adds more compatibility with no disavantages as you claim, then why not.

But, can't you just symlink /run/current-system/sw/bin/bash and /bin/bash ?

The-M1k3y:
Yes, I could just symlink the bash to /bin/bash, but that would counterfeit the whole concept of NixOS (since /run/current-system/sw/bin/bash itself is only a symlink).
But you are right, I forgot about the legacy distros. It's probably better to leave it like that for the moment and I just keep working with my fork.

dgibbs64:
Closing as no further action required currently

PR #1673, opened 2017-10-28
Added rp_downtown_v4c_v2

Doc1337:
I tested it and it is working fine.


UltimateByte:
Nope nope nope.
1) Maps are supposed to be added using the workshop so that clients download them off steam servers
2) This is a copy of a map not updated since 2012 that doesn't originate from the author since it was created 5h ago.

PR #1699, opened 2017-11-19
Changed the target-session of all tmux commands to use exact matching.

CasperAlant:
If tmux does not find target-session with an exact match, it will try to match the target-session parameter with the start of current session names. This pull-request simply forces tmux to use exact matching.

Source: http://manpages.ubuntu.com/manpages/zesty/man1/tmux.1.html

The problem manifests when running multiple servers (CS:GO in this example). The first session is named csgoserver and the second is named csgoserver-2. When using tmux ... -t "csgoserver", it will match csgoserver-2 if csgoserver is not running.

To recreate the problem:
```
wget -N --no-check-certificate https://gameservermanagers.com/dl/linuxgsm.sh && chmod +x linuxgsm.sh && bash linuxgsm.sh csgoserver
./csgoserver install
./linuxgsm.sh csgoserver
./csgoserver-2 install
./csgoserver start
./csgoserver-2 start
./csgoserver stop
```
Result:
This will stop **both** servers. csgoserver is stopped by fn_stop_graceful_cmd and csgoserver-2 is stopped by fn_stop_tmux.


dgibbs64:
Awesome thanks for the bug fix!!!

UltimateByte:
What does the `=` from `=${var}` do? Never seen this syntax.

CasperAlant:
From http://manpages.ubuntu.com/manpages/zesty/man1/tmux.1.html:

```
     target-session is tried as, in order:

           1.   A session ID prefixed with a $.

           2.   An exact name of a session (as listed by the list-sessions
                command).

           3.   The start of a session name, for example ‘mysess’ would match
                a session named ‘mysession’.

           4.   An fnmatch(3) pattern which is matched against the session
                name.

     If the session name is prefixed with an ‘=’, only an exact match is
     accepted (so ‘=mysess’ will only match exactly ‘mysess’, not
     ‘mysession’).

     If a single session is found, it is used as the target session; multiple
     matches produce an error.  If a session is omitted, the current session
     is used if available; if no current session is available, the most
     recently used is chosen.
```

UltimateByte:
Oh, so this is tmux specific, not really a bash thing. Thanks for satisfying my curiosity :D

Issue #1368, opened 2017-03-11
Checking dependencies: missing: python libstdc++.i686 


Richyro0:
Trying to solve an issue regarding launching an old Half-Life DM server. I get this message on install. 

Warning! hl2dmserver does not have sudo access. Manually install dependencies.
        sudo yum install python libstdc++.i686

When I try and do that i get the following...

Loaded plugins: priorities, update-motd, upgrade-helper
Package python26-2.6.9-2.88.amzn1.x86_64 already installed and latest version
No package libstdc++.i686 available.
Nothing to do

So I get No package libstdc++.i686 available.

Kind of hitting a wall here so I thought i'd log an issue. 

Thanks 

Rich

cedarlug:
You need to also tell us your distro - and a lot of other things.

Chances are you're running a Debian-based distro, and haven't followed the instructions to add the i38 architecture to your apt.sources file.

So that we don't have to guess any more, please follow the [posting guide](https://github.com/GameServerManagers/LinuxGSM/wiki/Support).  Bullet 4 is what I'm referring to above, but if you follow up, please include all the requested information.

Reopen if it's not a case of not following the directions on adding the architecture.

cedarlug:
Looking closer at the clues you gave, it appears to be an RPM-based distro (not Debian), but running on an amazon instance?

UltimateByte:
LGSM is a place for development.
Distro/user issues are solved on Discord or the Steam group, as per the mentioned guide.

gmwingard:
I think we may need to reopen this. It appears as if he was using amazon linux. They put the version in all of their packages, which the array depts won't be able to find (with how it's coded now). I've got a buddy that fixed it so that we can work with amazon linux. Should we get on discord to discuss?

PR #1702, opened 2017-11-22
Feature/issue 1368 - Amazon Linux Dependencies

tchawes:
This is regarding support for Amazon Linux.  All of the dependencies on latest Amazon Linux 64-bit AMI are pre-installed however they use version names in the repo so you must explicitly look for python27 and libstdc++64.i686.  Using the distroname variable I have added a case that looks for the appropriate packages for current Amazon Linux AMI's.  I have tested this on a t2.micro running the 7DTD installer with no issues. 

#1368


dgibbs64:
Nice!! I will merge soon :)

PR #1720, opened 2017-12-04
Create common-cfg

neuro6:
Commands that would apply to every or nearly every server in lgsm 

Commands that would typically be used or found in common.cfg

### DIFF
--- a/lgsm/config-default/config-lgsm/common-cfg
+++ b/lgsm/config-default/config-lgsm/common-cfg
@@ -1,0 +1,19 @@
+##################################
+######## Common Settings #########
+## Found in lgsm/config-lgsm/SERVERNAME/ ##
+##################################
+# PLACE GLOBAL SETTINGS HERE
+## These settings will apply to all instances
+
+# Set the max number of players that can connect
+## Most servers are 32, some are 24 or less
+## Un-Comment to change from the instance default
+# maxplayers=
+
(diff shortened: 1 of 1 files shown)


lgsm/config-default/config-lgsm/common-cfg
```diff
@@ -0,0 +1,19 @@
...
+## Un-Comment to change from the instance default
+# maxplayers=
+
+# Server exists behind a firewall NAT (does not have public IP)
+## Leave this as is unless you know better
+serverbehindnat=true
+
+# Set first map to load on server start
+# If this map does nto exist, then the server will not start correctly
+defaultmap="MAP_NAME"
```
MarkL4YG: I really like the idea of a common.cfg but I wouldn't really recommend having "defaultmap" as it sounds like a more specific option than the others. Some servers do not get their map names from their command line.

lgsm/config-default/config-lgsm/common-cfg
```diff
@@ -0,0 +1,19 @@
...
+## These settings will apply to all instances
+
+# Set the max number of players that can connect
+## Most servers are 32, some are 24 or less
+## Un-Comment to change from the instance default
+# maxplayers=
+
+# Server exists behind a firewall NAT (does not have public IP)
+## Leave this as is unless you know better
+serverbehindnat=true
```
dgibbs64: This var might be useful!

PR #1735, opened 2017-12-16
Update to support the 64bit binaries of Natural Selection 2

GhoulofGSG9:
Natural Selection 2 moved to 64bit a few month ago and doesn't ship the 32bit binaries anymore. Hence I updated the default config and the dependency and glibc checks.

Additionally NS2:Combat is still using a old 32bit fork of the spark engine, so we need to differ between NS2 and NS2:Combat when checking the dependencies.

Somebody would still need to update the dependencies listed at the lgsm ns2 install webpage. I couldn't find a repro for that.


dgibbs64:
Beautiful!. Thanks for the PR!

Issue #1775, opened 2018-01-16
Rust Oxide Mod downloading from old source


thedaedalus:
Hi 

I have been having issues with my Rust server and found that you are linking to an old Oxide repo. 

https://github.com/OxideMod/Oxide/releases

The new repo for Oxide for Rust is here

https://github.com/OxideMod/Oxide.Rust/releases

UltimateByte:
Ouch, thanks for the info. I'll update it ASAP.

UltimateByte:
Corrected. Run `update-lgsm` command to get the fix.

PR #1777, opened 2018-01-16
Updated oxide links using git api

UltimateByte:
Oxide links changed, they now use different repos. I just updated them using git API to find the latest link.


UltimateByte:
Fixes #1775

Issue #1752, opened 2017-12-27
Backup lock file included in backup


PyCactuar:
**Issue:**
When restoring a backup created with the (in this case) ./rustserver backup command, it also backs up the .backup.lock file located in /home/user/lgsm/tmp. After restoring the backup this causes issues when running a new backup task is started, as the lock file causes the system to believe a backup is already running.

**Possible Solution:**
If possible this could be solved by skipping the .lock.backup file when making a backup.

PyCactuar:
I've just located the following line in fn_backup_compression(): rm -f "${tmpdir}/.backup.lock"

For some reason this does not seem to happen when making a backup.

UltimateByte:
Hello,
Thanks for reporting.
I would rather solve this by excluding the tmpdir since it's useless to backup it anyways.

Issue #1727, opened 2017-12-08
realpath not found / wrong arguments when attempting backup on debian/ubuntu systems


smw:
Backup appears to be broken on ubuntu/debian systems, because `realpath` is not shipped with the coreutils on deb/ubuntu:   https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=693211.

There is a separate 'realpath' package, but that version doesn't appear to support the `--relative-to` argument.

This appears to be caused by this PR:  #1676 , which is a fix for this bug:    #1563 

Perhaps the usage of `realpath` could be replaced by `readlink -f` somehow?

```
steam@devi:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 14.04.5 LTS
Release:        14.04
Codename:       trusty
```

```
steam@devi:~$ ./sdtdserver backup
    fetching command_backup.sh...OK
[  OK  ] Backup sdtdserver: Backup starting
[ INFO ] Backup sdtdserver: There are no previous backups
[ .... ] Backup sdtdserver: Backup (3.3G) sdtdserver-2017-12-08-000441.tar.gz, in progress.../home/steam/lgsm/functions/command_backup.sh: line 115: realpath: command not found
tar: ./lgsm/backup/sdtdserver-2017-12-08-000441.tar.gz: file changed as we read it
FAIL
[ FAIL ] Backup sdtdserver: Starting backup
steam@devi:~$ realpath
The program 'realpath' is currently not installed. To run 'realpath' please ask your administrator to install the package 'realpath'
```

```
root@devi:~# apt install realpath
[snip]
```

```
steam@devi:~$ dpkg -l realpath
[snip]
ii  realpath                              1.19                    amd64                   Return the canonicalized absolute pathname
```
```
steam@devi:~$ ./sdtdserver backup
[  OK  ] Backup sdtdserver: Backup starting
        * Previous backup was created less than 1 day ago, total size 4.5G
[ WARN ] Backup sdtdserver: sdtdserver will be stopped during the backup
[  OK  ] Stopping sdtdserver: Graceful: telnet: 127.0.0.1: OK
[  OK  ] Stopping sdtdserver: devi
[ .... ] Backup sdtdserver: Backup (3.3G) sdtdserver-2017-12-08-013330.tar.gz, in progress...realpath: unrecognized option '--relative-to=/home/steam'
Usage:
[...]

PR #1786, opened 2018-01-23
relative path using readlink instead of realpath

cedarlug:
Adding a function to bypass the idiosyncrasies of different realpath implementations which use different syntax.

Also adds an exclusion for backing up the lockfile in the backup tarball.
Ref: #1727, #1752


UltimateByte:
@cedarlug Is this tested working?
If so, we should push it to master, because backup is still broken...

cedarlug:
It has been working for me as expected. 

However I don't hammer on backups like others might because I'm using a full system backup/restore infrastructure.

UltimateByte:
Well, I don't see how it could be worst than a backup backuping itself.

Issue #1664, opened 2017-10-21
Discord Notification: Don't use private IP in discord notification


robobenklein:
https://github.com/GameServerManagers/LinuxGSM/blob/8b163a61c44f237ffb45fcb9cc4cd9cc237a73fd/lgsm/functions/alert_discord.sh#L42

Test environment: TF2 on Ubuntu server.

When on a server with multiple interfaces, we need to set the `ip` variable so that the source engine knows where to bind the port, but that's a local network IP address. (192.168.X.X or whatever the network might be)

In the discord notification, it then uses the private IP address, meaning it's useless to have a gameserver tracker link there at all; nobody but the ones inside the network can even know what the address is.

## Proposed workaround:
Allow custom variable to override `ip` variable.

Changes here:
```
"value": "[${ip}:${port}](https://www.gametracker.com/server_info/${ip}:${port})"
```
to
```
"value": "[${public_ip:-$ip}:${public_port:-$port}](https://www.gametracker.com/server_info/${public_ip:-$ip}:${public_port:-$port})"
```

Using `public_ip` and `public_port` if given allowing `ip` and `port` to be overridden if the router or network changes them in transit to the server.

I could submit a PR if needed, but this is such a small change I figured it wasn't that big a deal.

dgibbs64:
I like this idea. Thanks

gamersalpha:
same problem 

thanks

dgibbs64:
LinuxGSM now has an internet ip variable. I think I will set it to this instead

dgibbs64:
I will be altering the IP to use extip however some though does need to go in to all the possible variations of how IP addresses may change

dgibbs64:
Updated this for other notifications as well

PR #1787, opened 2018-01-24
Discord alert: Allow public_ ip and port overrides

robobenklein:
https://github.com/GameServerManagers/LinuxGSM/issues/1664

### DIFF
--- a/lgsm/functions/alert_discord.sh
+++ b/lgsm/functions/alert_discord.sh
@@ -39,7 +39,7 @@ json=$(cat <<EOF
 			},
 			{
 				"name": "Server IP",
-				"value": "[${ip}:${port}](https://www.gametracker.com/server_info/${ip}:${port})"
+				"value": "[${public_ip:-$ip}:${public_port:-$port}](https://www.gametracker.com/server_info/${public_ip:-$ip}:${public_port:-$port})"
 			},
 			{
 				"name": "More info",


UltimateByte:
Good idea, but it won't work if public IP is an IPv6, as it currently occurs when ipv6 is available.

robobenklein:
@UltimateByte Can you explain further? Perhaps give an example?

This doesn't introduce any new bugs as far as I'm aware, it just gives another option for resolving a case in which the IP field of the notification is wrong.

UltimateByte:
Quick example : 

````bash
root@gaben:~# public_ip="10:AB:24"; ip="127.0.0.1"; value="${public_ip:-ip}"; echo "$value"
10:AB:24
````

Well in my case ( https://hastebin.com/osefopepay ) it would try to query the game server on ipv6, which isn't supported.

I'm not saying your code is bad, but in order to not introduce a bug, you would need to check if the public IP is an IPv6 or not. 

Since I don't think any game server supports IPv6, or at least not many as of now, the best solution would be to correct the IP checking system to show only public IPv4.

robobenklein:
This change *only* modifies the IP given through a Discord notification, it doesn't change anything about how LGSM queries a server or binds to an IP. You could set this `$public_ip` to literally anything you want and it wouldn't affect your use case at all unless you were using Discord alerts.

Here the `$public_ip` is not going to be used in LGSM at all, it's purely for humans to see.

For example, I plan to set mine to something akin to `tf2ware.robosane.net` which isn't an IP at all. I could even set `public_ip="lolsomething that's not an ip"` and the only problem would be that the GameTracker.com URL in the Discord notif wouldn't work...

robobenklein:
Actually @UltimateByte you just helped me realize how confusing this is calling it `public_ip`. I'm gonna redo the discord notification thing to make it much easier and simpler to configure.

PR #1788, opened 2018-01-25
Discord Alert customization

robobenklein:
Adds the following overriding options for discord alerts:

```
discord_bot_name: Name shown for the bot message
discord_avatar_url: Direct link to image to be used for icon & avatar in message
discord_public_ip: Overrides $ip for notif if set
discord_public_port: Overrides $port for notif if set
```

### DIFF
--- a/lgsm/functions/alert_discord.sh
+++ b/lgsm/functions/alert_discord.sh
@@ -39,7 +39,7 @@ json=$(cat <<EOF
 			},
 			{
 				"name": "Server IP",
-				"value": "[${ip}:${port}](https://www.gametracker.com/server_info/${ip}:${port})"
+				"value": "[${public_ip:-$ip}:${public_port:-$port}](https://www.gametracker.com/server_info/${public_ip:-$ip}:${public_port:-$port})"
 			},
 			{
 				"name": "More info",
--- a/lgsm/functions/alert_discord.sh
+++ b/lgsm/functions/alert_discord.sh
@@ -5,21 +5,26 @@
 # Website: https://gameservermanagers.com
 # Description: Sends Discord alert.
 
+bot_name="${discord_bot_name:-LinuxGSM}"
+avatar_url="${discord_avatar_url:-https://raw.githubusercontent.com/GameServerManagers/LinuxGSM/master/images/logo/lgsm-dark-square-512.png}"
+shown_ip_address="${discord_public_ip:-$ip}"
+shown_port="${discord_public_port:-$port}"
+
 json=$(cat <<EOF
 {
-"username":"LinuxGSM",
-"avatar_url":"https://raw.githubusercontent.com/GameServerManagers/LinuxGSM/master/images/logo/lgsm-dark-square-512.png",
+"username":"${bot_name}",
+"avatar_url":"${avatar_url}",
 "file":"content",
 
 "embeds": [{
 	"color": "2067276",
-	"author": {"name": "${alertemoji} ${alertsubject} ${alertemoji}", "icon_url": "https://raw.githubusercontent.com/GameServerManagers/LinuxGSM/master/images/logo/lgsm-dark-square-512.png"},
+	"author": {"name": "${alertemoji} ${alertsubject} ${alertemoji}", "icon_url": "${avatar_url}"},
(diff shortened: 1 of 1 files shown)


UltimateByte:
I hope you won't hate me too much for this, but I still have major issues with your code.

1. I still don't like `discord_public_ip` , because LinuxGSM is designed for the user to only have one IP to input: his listenning address.
Public IP can be detected by another LinuxGSM's function, the one that already tries to do that can be improved. The complicated case might be when multiple public IPs are listening, or detecting if this main IP address that is set is public or local.

2. LinuxGSM code should not be edited by the user, ever, because it would be resetted upon any LGSM update if customized. Therefore, it is pointless to add these settings inside a script file since it breaks updates for the user.
So these settings should be added inside settings files, while respecting LinuxGSM's variable convention: all lowercase with no signs variables as per: https://github.com/GameServerManagers/LinuxGSM/wiki/Syntax-and-Conventions#naming-variables

Note: You don't need to start over with a new PR, you can edit your repo, it will add any change to the PR.

robobenklein:
Here's why I used `discord_public_ip` instead of doing detection or a function, domain names.

I wanted to be able to set the "IP" to be the full dns name for the IP, instead of using a public ipv4, the only other way to be able to detect a domain name instead of the public IP would be to look at the FQDN of the host, which is much harder (sometimes impossible) to configure and set properly on most providers. Even then, in my use case I am using a CNAME subdomain to point to the FQDN of the host, making it almost impossible (completely unreasonable) for automatic detection to be used here.

As for point 2, I don't really see how that's relevant. The user won't be editing this file at all, they will set those 4 `discord_` variables in their custom or server-specific configs. All the variables are snake-cased with underscores, and the temporary variables like `shown_port` are all scoped to the alert_discord script.

The way it's implemented in this PR does not affect existing functionality at all, doesn't introduce any new bugs, and adds a feature set which originally could only be accomplished by doing what you said, editing the function scripts.

If the user doesn't use discord alerts at all, it doesn't affect them at all.

I don't see any downsides to these changes, what would you do?

UltimateByte:
Lol, maybe I understood your code the wrong way then.
Where is `discord_bot_name` defined for example?

robobenklein:
For example,
```
tf2server@trurl:~$ cat lgsm/config-lgsm/tf2server/tf2server.cfg 
##################################
######## Instance Settings ########
##################################
# PLACE INSTANCE SETTINGS HERE
## These settings will apply to a specific instance

#ip="tf2ware.robosane.net"
ip="192.168.10.67"
maxplayers="24"
defaultmap="tf2ware2_a4"
mport="27015"

gslt="52XXXXXXXXXXXXX92"

# Discord Alert
# Discord Alerts | https://github.com/GameServerManagers/LinuxGSM/wiki/Discord
discordalert="on"
discordwebhook="https://canary.discordapp.com/api/webhooks/XXX/XXX"
discord_public_ip="tf2ware.robosane.net"
discord_bot_name="Robosane TF2ware"
discord_avatar_url="https://steamcdn-a.akamaihd.net/steamcommunity/public/images/avatars/b7/b7225dda3950f425031975b19d1e371172b45159_full.jpg"
```
image

And clicking on the url (https://www.gametracker.com/server_info/tf2ware.robosane.net:27015) still works because it resolves the domain to the IP.

UltimateByte:
Thanks, now I think I get the full thing.

So:
1) As I already said, these vars don't fit the naming convention : text only, no signs, no caps as per https://github.com/GameServerManagers/LinuxGSM/wiki/Syntax-and-Conventions#naming-variables
2) These are config vars, therefore, they should be added to config files https://github.com/GameServerManagers/LinuxGSM/tree/master/lgsm/config-default/config-lgsm
3) For the public IP thing, I see what you're trying to do here, it's rather cosmetics than functional. I think you can go further with your mechanics and make it a global thing for LGSM, a var like `displayedserveraddress` that could be used anywhere the LGSM's public IP is needed. It could for example be added to `details` to other notification systems.

UltimateByte:
Not updated for a month, closing.
This feature would be welcome, but would need to fit the upper requirements (at least 1 and 2).

PR #1824, opened 2018-02-26
Update _default.cfg

Khelgar1:
Replacing the steam login details with an anonymous login will allow the install of the 7dtd server to complete successfully without having to edit common.cfg beforehand (which is only created when you run ./sdtdserver install which fails). The webpage information is misleading currently, telling you to edit a file that doesn't yet exist.

### DIFF
--- a/lgsm/config-default/config-lgsm/sdtdserver/_default.cfg
+++ b/lgsm/config-default/config-lgsm/sdtdserver/_default.cfg
@@ -9,8 +9,8 @@
 #### Server Settings ####
 
 ## SteamCMD Login | https://github.com/GameServerManagers/LinuxGSM/wiki/SteamCMD#steamcmd-login
-steamuser="username"
-steampass='password'
+steamuser="anonymous"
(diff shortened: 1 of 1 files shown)


UltimateByte:
Your fix is wrong in the sense that `anonymous` is supposed to be used when no username is defined and that the game is known to use anonymous login. So if sdtd doesn't switch to anonymous as expected, it needs to be fixed at the source. Can't remember where this is defined, some search is needed.

Khelgar1:
Ahh I didn't know it was supposed to do that. Maybe 7dtd isn't  flagged as a server that can use anonymous?
My fix would work though,  but shouldn't be necessary?

UltimateByte:
Well, I just got that default values are:
steamuser="username"
steampass='password'

So if it is now possible to download the game as anonymous, the correct fix is to just remove these variables, then it will use anonymous since these vars will be undefined.

Khelgar1:
I'll close this so and let you just remove those?

PR #1843, opened 2018-03-14
Adding debugging information

cedarlug:
Digging into travis issues


cedarlug:
Clearing out - can't seem to remove prior cruft.

Issue #1839, opened 2018-03-12
Travis-CI failing because of changes to backup realpath


dgibbs64:
@cedarlug please could you investigate the source of buids failure in Travis-CI. It relates to your improvement of backups https://github.com/GameServerManagers/LinuxGSM/pull/1786 

error is.

    [ INFO ] Backup tests_jc2server.sh: Problem identifying the previous backup directory for exclusion.

cedarlug:
Will dive in...

dgibbs64:
#1845

dgibbs64:
connects to #1845

h3o66:
Closing - becasue moving to github actions

PR #1845, opened 2018-03-15
Travis testing with the relpath alternative #1839

cedarlug:
Working through issues suspected as being caused by first-run omission of an existing backup directory.

connects to #1839


dgibbs64:
Any luck with this? Could the failure be something to do with travis running ubuntu 16.04? I will bypass the backup test until this is fixed

cedarlug:
I believe it's just an over-zealous error check in my code that's barfing when backup is run the first time, because the "exclude" directory doesn't yet exist, but will be created by the first rsync.

Trying to confirm this hypothesis has been a bit problematic.  Cross fingers, revisiting presently.

cedarlug:
Any insight on why there suddenly exists a lockfile where none existed previously would be greatly appreciated.

dgibbs64:
There has always been a lock file that is created when a backup runs.

cedarlug:
The crux of the issue is that the travis build fails at step 7, the backup, with the existence of a lockfile that shouldn't be there. 

Further if I remove that lockfile check, travis still dies with the same error.  Pushing new commit to illustrate:

cedarlug:
What's been happening is that the travis build dies at step 7 with the backup check on the existence of the lockfile with the phrase "Lock file found: Backup already in progress."  Even if I completely remove that check and that specific line of prose from the command_backup.sh script.

Hoping that it works as expected now, but that's what's been happening.

cedarlug:
That failed again.  Here's what I'm seeing:

- The build fails with "Lock file found: Backup is currently running", but this phrase IS NOWHERE in the current codebase (see the cumulative patch -It's replaced with "happy thoughts").
`cedarlug@mi6-scanner:~/LinuxGSM$ grep -d recurse "Lock file found:" .
./lgsm/functions/command_backup.sh:             fn_print_info_nl "Lock file found: Think happy thoughts"
`
- The patch for the pull request shouldn't die at this check anyway - just return happy thoughts in an info.

The cumulative patch is small - the goal was to test the insertion of the 5 lines checking to see if the target backup didn't exist yet while leaving in the sanity check on the return of relpath result.  Click on the "Files changed (1)" above to see the diff showing the intended check and (now) the removal of the lockfile check that is still killing the travis build - even though it's been removed.  <--- My current nemesis.

In order to check to see if it was simply my travis environment that was corrupted, I created a new github user (you may have seen the 'totallynotcedarlug' pull requests), and the same happens with a brand new user.

dgibbs64:
Seems like a right pain in the arse. Thanks for the update.

UltimateByte:
So Travis is not getting/using latest files in some way...?

Just a guess, but what about `githubbranch="$TRAVIS_BRANCH"` (used here https://github.com/GameServerManagers/LinuxGSM/blob/master/tests/tests_ts3server.sh#L35 ) does not point to what we think?

It seems like documentation needed updates, but change were not comitted: https://github.com/travis-ci/travis-ci/issues/8200

https://github.com/Junkern/docs-travis-ci-com/blob/8346ceb11cf0d87390611e6a4aafcb26163c2288/user/environment-variables.md#default-environment-variables

> TRAVIS_BRANCH:
> for push builds, or builds not triggered by a pull request, this is the name of the branch.
> for builds triggered by a pull request this is the name of the branch targeted by the pull request.
> for builds triggered by a tag, this is the same as the name of the tag (TRAVIS_TAG). This is because,

cedarlug:
Looks like TRAVIS_BRANCH is expanding to "master" in the travis output.

dgibbs64:
connects to #1839

h3o66:
Closing - becasue moving to github actions

PR #1874, opened 2018-04-11
Fix support bot config

dessant:
Got a Sentry notification for the bot instance with this error:
```
incomplete explicit mapping pair; a key node is missed; or followed by a non-tabulated empty line at line 4, column 19:
    supportLabel: Type: support
```


dgibbs64:
thank you

PR #1902, opened 2018-05-17
Update alert_telegram.sh

lazymind:
I'm add a simple curl custom string for use curl with proxy in countries like Russia where Telegram has blocked.


UltimateByte:
You should add this to your previous PR. No need to make 1 PR per commit. 
Also, Telegram is used for all game servers, so the var should be added to other _default.cfg as well.

PR #1952, opened 2018-07-21
Fixed wrong text in some core messages

UltimateByte:
This has probably been here for a while, but truth is... we don't see them quite often :D


dgibbs64:
im going to merge this in to develop as it is not a software breaking bug

PR #1953, opened 2018-07-21
Hotfix/minecraft 

### LINKED ISSUE #1944: [mcserver] Latest update does not download
A new version of Minecraft is available however something has changed meaning the download does not work currently. I will investigate

UltimateByte:
fixes #1944

Updated downloads URLs using jq to get latest minecraft server versions
Also corrected that "nogui" nonsense thing that was was requested, which breaks the executable check function. Also, it is already in parms so it is applied already and there should be no need for it there.

### DIFF
--- a/lgsm/config-default/config-lgsm/mcserver/_default.cfg
+++ b/lgsm/config-default/config-lgsm/mcserver/_default.cfg
@@ -89,7 +89,7 @@ engine="lwjgl2"
 ## Server Specific Directories
 systemdir="${serverfiles}"
 executabledir="${serverfiles}"
-executable="java -Xmx${javaram}M -jar ${serverfiles}/minecraft_server.jar nogui"
+executable="java -Xmx${javaram}M -jar ${serverfiles}/minecraft_server.jar"
 servercfg="server.properties"
 servercfgdefault="server.properties"
 servercfgdir="${serverfiles}"
--- a/lgsm/functions/update_minecraft.sh
+++ b/lgsm/functions/update_minecraft.sh
@@ -9,7 +9,8 @@ local commandaction="Update"
 local function_selfname="$(basename "$(readlink -f "${BASH_SOURCE[0]}")")"
 
 fn_update_dl(){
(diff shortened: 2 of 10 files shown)


lgsm/config-default/config-lgsm/mcserver/_default.cfg
```diff
@@ -89,7 +89,7 @@ engine="lwjgl2"
 ## Server Specific Directories
 systemdir="${serverfiles}"
 executabledir="${serverfiles}"
-executable="java -Xmx${javaram}M -jar ${serverfiles}/minecraft_server.jar nogui"
```
dgibbs64: `nogui` was required so would need to be in parameters instead.

dgibbs64:
jq will need to be added as a standard lgsm dependency now

lgsm/functions/update_minecraft.sh
```diff
@@ -69,7 +71,7 @@ fn_update_currentbuild(){
 
 fn_update_availablebuild(){
 	# Gets latest build info.
-	availablebuild=$(${curlpath} -s "https://launchermeta.mojang.com/mc/game/version_manifest.json" | sed -e 's/^.*"release":"\([^"]*\)".*$/\1/')
+	availablebuild=$(${curlpath} -s "http://launchermeta.mojang.com/mc/game/version_manifest.json" | jq -r '.latest.release')
```
dgibbs64: please change to `https://`

lgsm/config-default/config-lgsm/mcserver/_default.cfg
UltimateByte: nogui is already in fn_parms, so the the result should be identical to this : added right after the executable.
My testing showed that linuxgsm would fail to find the executable with "nogui" added here, so any case, it's better removed, and that it also works perfectly fine without it.

dgibbs64: o yeah I missed that

lgsm/functions/update_minecraft.sh
UltimateByte: Done!

dgibbs64:
Had to do some funky stuff not now merged in to master

Issue #1847, opened 2018-03-15
Is server listed on the Steam master server


dgibbs64:
Using the steam API it is possible to confirm if a server is listed on there master server. It would be very easy to add this to details to allow admins to confirm if the server is being listed on the master server. Could also be used as part of a monitor feature to notify if a serve is not listed.

http://api.steampowered.com/ISteamApps/GetServersAtAddress/v0001?addr=216.52.148.47:27015&format=json

dgibbs64:
connects to #1986

PR #1986, opened 2018-08-12
Details will now list if game server is on the Steam Master server

dgibbs64:
Applies to servers that are on the steam master server.

* Source Engine
* Gold Source Enginer
* Just Cause 2
* Quake Live

If there are any missing please let us know


dgibbs64:
connects to #1847

Issue #1555, opened 2017-07-31
lightgreen not defined/printf does not like colours


drphlux:
- [x] lightgreen is not defined
- [x] printf does not like colours

Example: 
```
[sdtdserver@mynuc ~]$ grep lightgreen lgsm/functions/command_details.sh 
	echo -e "${lightgreen}${gamename} Server Details${default}"
	echo -e "${lightgreen}${selfname} Script Details${default}"
	echo -e "${lightgreen}Backups${default}"
	echo -e "${lightgreen}Command-line Parameters${default}"
	echo -e "${lightgreen}Ports${default}"
	echo -e "${lightgreen}${servername} WebAdmin${default}"
	echo -e "${lightgreen}${servername} Telnet${default}"
	echo -e "${lightgreen}${servername} WebAdmin${default}"
	echo -e "${lightgreen}${servername} WebAdmin${default}"
```
But lightgreen is not defined:
```
[sdtdserver@mynuc ~]$ grep light lgsm/functions/core_messages.sh 
	lightyellow="\e[93m"
	lightblue="\e[94m"
```
Also, printf seems to mess up the color coding: 

```
Performance
========================
\x1b[34mUptime:    \x1b[0m6d, 6h, 39m
\x1b[34mAvg Load:  \x1b[0m0.45, 0.62, 0.73

\x1b[34mMem:       \x1b[34mtotal   used   free        cached\x1b[0m
\x1b[34mPhysical:  \x1b[0m31G     1.9G   16G         47M\x1b[0m
\x1b[34mSwap:      \x1b[0m15G     0B     15G\x1b[0m  
```

smounives:
same here.

PR #1987, opened 2018-08-12
Added missing Colours

dgibbs64:
Added missing ansi colours and added an option to disabled ansi colours all together


dgibbs64:
connects to #1555

Issue #1955, opened 2018-07-21
Minecraft Query does not work


dgibbs64:
**Describe the bug**
Minecraft querying does not work currently

**To Reproduce**
Try `./mcserver query-raw`

**Expected behaviour**
Minecraft querying should work and has been tested as working

PR #2003, opened 2018-09-08
Added Query support for Minecraft #1955

dgibbs64:
* Added Query support for Minecraft
* Fixed query enabled display in details


dgibbs64:
connects to #1955

PR #2008, opened 2018-09-11
Fix OxideRust download link

MrGraversen:
The OxideMod/Oxide.Rust GitHub repository is now found under theumod/umod.rust.

This, of course, has caused Oxide update to fail, and is simply fixed by swapping the URL. The Hurtworld and SevenDaysToDie are still belonging to the OxideMod account, but it may also migrate later.

See: https://github.com/theumod/umod.rust

### DIFF
--- a/lgsm/functions/mods_list.sh
+++ b/lgsm/functions/mods_list.sh
@@ -27,7 +27,7 @@ sourcemodlatestfile="$(wget "${sourcemodscrapeurl}" -q -O -)"
 sourcemoddownloadurl="https://www.sourcemod.net/latest.php?os=linux&version=${sourcemodmversion}"
 sourcemodurl="${sourcemoddownloadurl}"
 # Oxide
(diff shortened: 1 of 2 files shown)


MrGraversen:
I verified this by swapping the download link in `mods_list.sh` on one of my servers and performing `bash rustserver mu`.

UltimateByte:
Perfect, that's the info that I needed.
Thanks for the PR, should be merged soon.
Ref:
fixes #2011

MrGraversen:
Great, thanks!

dgibbs64:
This can be a hotfix

lukespragg:
I'd suggest having Curl follow redirects, as the old URL would still work as well.

-oxiderustlatestlink="$(curl -L -s https://api.github.com/repos/OxideMod/Oxide.Rust/releases/latest | grep browser_download_url | cut -d '"' -f 4)"

https://davidwalsh.name/curl-follow-redirects

dgibbs64:
connects to #2011

PR #2019, opened 2018-09-27
Fix Ctrl+C closing wrong session

### LINKED ISSUE #2018: Server stop with Ctrl+C causes wrong server to close
**Describe the bug**
Stopping a server which requires a Ctrl+C in tmux causes random servers to close instead of the target server. 

**To Reproduce**
Steps to reproduce the behaviour:
1) Launch 2 KF2 dedicated servers
2) Stop the first dedicated server launched
3) Watch as both servers close

**Expected behaviour**
Only the target server is closed

The cause of this is due to [command_stop.sh Line 17](https://github.com/Bourne-ID/LinuxGSM/blob/develop/lgsm/functions/command_stop.sh#L17) which has the arguments in the wrong order, therefore sending the command 'Ctrl+C -t "[server name]"' instead of Ctrl+C to a target session.

Bourne-ID:
Fixes #2018


dgibbs64:
the pull request just failed. I am releasing this as a hotfix. Please, can you recreate the PR from the master branch and I will merge as a hotfix

PR #2026, opened 2018-10-05
Add fn_start_docker

ptrj:
Docker support. Start game without tmux and game runs in foreground.
[Why without tmux?](https://gist.github.com/ptrj/b54b8cdd34f632028e669a1e71bc03b8)
Hope it helps to push forward your [docker repository](https://github.com/GameServerManagers/LinuxGSM-Docker).

### DIFF
--- a/lgsm/functions/command_start.sh
+++ b/lgsm/functions/command_start.sh
@@ -187,6 +187,28 @@ sleep 0.5
 	echo -en "\n"
 }
 
+fn_start_docker(){
+	fn_parms
+	# Log rotation
+	fn_script_log_info "Rotating log files"
+	if [ "${engine}" == "unreal2" ]&&[ -f "${gamelog}" ]; then
+		mv "${gamelog}" "${gamelogdate}"
+	fi
+	if [ -f "${lgsmlog}" ]; then
+		mv "${lgsmlog}" "${lgsmlogdate}"
+	fi
+	if [ -f "${consolelog}" ]; then
+		mv "${consolelog}" "${consolelogdate}"
+	fi
+	# Create logfile
+	touch "${consolelog}"
+	
+	# Create lockfile
+	date > "${rootdir}/${lockselfname}"
+	cd "${executabledir}"
+	${executable} ${parms} 2> ${lgsmlogdir}/.${servicename}-tmux-error.tmp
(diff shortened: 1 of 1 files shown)


dgibbs64:
This is great thank you. My docker knowledge is very limited so I am relying on more experienced docker developers to help. 

Am i right tin saying that tmux doesn't run well in docker? if so it might mean that I treat all servers the same as teamspeak3 (no tmux)

ptrj:
You are welcome. I'm not a docker guru, I'm a fan of LinuxGSM and I want to use the Docker in future. 
Looks like tmux needs terminal inside Docker container. Argument "-t"
```
docker run -t -d image/name:version
```
Example above works, but if I use [docker-compose](https://docs.docker.com/compose/) or [Docker orchestration](https://www.docker.com/products/orchestration) like a Kubernetes, then it will not work.
Best practice for Docker container is one process running in foreground. If this process ends, then the container will terminate as well.
Then my opinion is, don't use tmux in container.
Oh, and when you speak about teamspeak, i forgot add some lines in `command_start.sh`
```
        if [ "${ts3serverpass}" == "1" ]; then
                if [ "${docker}" == "on" ]; then
                        ./ts3server_minimal_runscript.sh start serveradmin_password="${newpassword}" inifile="${servercfgfullpath}"
                else
                        ./ts3server_startscript.sh start serveradmin_password="${newpassword}" inifile="${servercfgfullpath}" > /dev/null 2>&1
                fi
        else
                if [ "${docker}" == "on" ]; then
                        ./ts3server_minimal_runscript.sh start inifile="${servercfgfullpath}"
                else
                        ./ts3server_startscript.sh start inifile="${servercfgfullpath}" > /dev/null 2>&1
                fi
        fi
```
Thank you

dgibbs64:
Awesome! So basically treat docker almost as a remix replacement and that will be the best way to do it.

Roni1993:
Hi @dgibbs64, I'm currently fiddling around with a docker-based setup for LGSM and I was stumbling upon this.
Is there any way I can support you with this PR?

ptrj:
Hi @dgibbs64,
I think you can close this RP, somewhere over the years to make a workable solution :-)

PR #2030, opened 2018-10-07
Created CPU_Protect

AkhilVenu:
This shell script willl stop all the process  that increases the cpu utilization over  a particular limit.


ptrj:
This script try to kill every process which use more then 70% CPU, not only game server. Why not use the "nice" command?

dgibbs64:
Does not have any relevance to LinuxGSM closing PR

PR #2032, opened 2018-10-07
Dependency checking can run as root

dgibbs64:
dependency checks can now run as root in "server mode" e.g `./csgoserver install` run as root will trigger dependency checking only. Allowing users to install dependencies using LinuxGSM. This will be very useful for automated installations of LinuxGSM


UltimateByte:
Huh, I hope your worked on that one a lot, because it can become a mess real quick.

PR #2034, opened 2018-10-09
Added TTT as CS:GO Gamemode

Bara:
I tested it on my (ttt) test server and works fine. Let me know if anyone has a other opinion.


UltimateByte:
Why not? Thanks for the submit :)

Issue #1460, opened 2017-06-04
[Server Request] Wurm Unlimited


ghost:
requesting a wurm unlimited LGSM server. I am willing to create this, I just need some guidance.

UltimateByte:
Guidance is on the wiki if you want to go through this on your own.
Though it should be noted that LinuxGSM team is currently focused on reworking the main code.

https://github.com/GameServerManagers/LinuxGSM/wiki/Developing-LGSM
https://github.com/GameServerManagers/LinuxGSM/wiki/Development-Workflow
https://github.com/GameServerManagers/LinuxGSM/wiki/Syntax-%26-Conventions
https://github.com/GameServerManagers/LinuxGSM/wiki/LGSM-Exit-Codes
https://github.com/GameServerManagers/LinuxGSM/wiki/Branching
https://github.com/GameServerManagers/LinuxGSM/wiki/Developer-Commands

Ultimately, any proper server request should come with server documentation, unless you're 100% sure you can handle everything yourself.

Found this
http://www.wurmpedia.com/index.php/Server_administration_(Wurm_Unlimited)

dgibbs64:
User on Discord requested this server today

interstate420productions:
Best tutorial I have found to date. I got my server up and running ONE time with this tutorial.  I can't get it to go again... :(  Frustrating.

dgibbs64:
I could really do with a poll to find out which servers people want the most

interstate420productions:
easy enough plenty of free ones on the web that are supposedly easy to embed in a site

interstate420productions:
I found this for you perhaps this will be of some use... it's supposed to be a completely automated script.. But it's not working for me.   https://github.com/jonsjava/wurm_unlimited_server

PR #2041, opened 2018-10-14
Wurm Unlimited

dgibbs64:
Wurm Unlimited server is now ready
* debug does not work
* gamedig is unavailable but gsquery does work


dgibbs64:
closes #1460

PR #2046, opened 2018-10-16
Design tweaks and JSON-safe escaping string

diamondburned:
A lot of people have been getting a "Cannot send an empty message", as their hostname contains characters that weren't JSON-safe (as in you can't literally just throw the variable into a pre-made JSON body and ship it away). 

Since `jq` was made into a forced dependency (which I'm not happy about, since only one game u-), I've decided to put it to good use.

### DIFF
--- a/lgsm/functions/alert_discord.sh
+++ b/lgsm/functions/alert_discord.sh
@@ -54,8 +54,11 @@ EOF
 
 fn_print_dots "Sending Discord alert"
 
-minified="$(echo -n "$json" | jq -c -M "@json")"
-discordsend=$(${curlpath} -sSL -H "Content-Type: application/json" -X POST -d "${json}" "${discordwebhook}")
+minified="$(echo -n "$json" | jq -c .)"
+
+echo "$minified" > /tmp/json
+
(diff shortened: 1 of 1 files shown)


dgibbs64:
Cool. We are finding new uses for jq. Used it to display if a server is on the steam master server as well.

lgsm/functions/alert_discord.sh
```diff
@@ -5,60 +5,65 @@
 # Website: https://linuxgsm.com
 # Description: Sends Discord alert.
 
+if ! command -v jq > /dev/null; then
+    fn_print_fail_nl "Sending Discord alert: jq is missing."
+    fn_script_log_fatal "Sending Discord alert: jq is missing."
+fi
+
+EscapedServername="$(echo -n "${servername}" | jq -sRr "@json")"
```
Bourne-ID: Do we use camel case for variable names elsewhere in the project?

diamondburned: Nope, just my habit. I'll push a commit tomorrow if I have time as it's late now.

diamondburned:
Made a commit to fit Camel Case requirements + removed extra debug lines.

diamondburned:
Changed, although I'm not sure if it'll be JSON-friendly or not (`jq` might have that handled though)

Issue #2062, opened 2018-10-29
TF2 server segfaults on startup in Fedora 28 x86_64


vectorsigma:
**Describe the bug**
In Fedora 28, for Team Fortress 2, `replay_srv.so` is unable to load, causing segfault and server crash loop in the console.  The server will refuse to load.

**To Reproduce**
Steps to reproduce the behaviour:
1. download lgsm for tf2 and follow [installation instructions](https://linuxgsm.com/lgsm/tf2server/) in the home directory of a newly created `tf2` user.
2. execute `./tf2server start`
3. execute `./tf2serer console`
4. press enter at the molly guard warning
5. See the error snippet below:
```
server_srv.so loaded for "Team Fortress"
Could not load: replay_srv.so
Could not load: replay_srv.so

Looking up breakpad interfaces from steamclient
Calling BreakpadMiniDumpSystemInit
./srcds_run: line 324:  2506 Segmentation fault      (core dumped) $HL_CMD
Add "-debug" to the ./srcds_run command line to generate a debug.log to help with solving this problem
Sat Oct 27 14:03:29 PDT 2018: Server restart in 10 seconds
```

**Expected behaviour**
Server software starts without crashing.

**Minimum Information**
Running the TF2 game, on Fedora 28, x86_64.  Installed fresh on 2018/10/27.

**Additional context**
Turns out, this is just a simple bug because `replay_srv.so` is compiled to expect the library to be named `libcurl-gnutls.so.4`, which the Fedora 28 version of `libcurl.i686` doesn't provide, by that exact name.  You can solve it as follows: `ln -nfs /usr/lib/libcurl.so.4 ~/serverfiles/bin/libcurl-gnutls.so.4` as the user you installed the server as.

Proposed fix in #2063.

tkrn:
On RHEL 7.5, I had the same issue. The following fix resolved the segfault.

`ln -s /usr/lib/libcurl.so.4 /home/tf2server/serverfiles/bin/libcurl-gnutls.so.4`

After this, it resolved the issue. Please note that on RHEL 7.5 libcurl.so.4 is a default symlink to  libcurl.so.4.3.0 which is provided in the 'libcurl' package in the yum repo.

```
[tkrn@lanserv:~]$ls -lah /usr/lib | grep libcurl
lrwxrwxrwx.  1 root root   16 Jul 30 14:05 libcurl.so.4 -> libcurl.so.4.3.0
[...]

PR #2063, opened 2018-10-29
Fix the missing library segfault issue.

vectorsigma:
This should resolve the TF2 segfault issue in `replay_srv.so`  for #2062


dgibbs64:
Thanks for this. An identiclt fix located in fix_dst.sh exists. I will merge this then alter the fix to match fix_dst.sh. Thanks for submitting. 

Going in to hotfix release today as important

vectorsigma:
> Thanks for this. An identical fix located in fix_dst.sh exists. I will merge this then alter the fix to match fix_dst.sh. Thanks for submitting.

You're welcome, and thanks for LinuxGSM! :+1:

PR #2067, opened 2018-10-31
Add some default mapgroup examples to csgoserver

borzaka:
There is no default specific mapgroup for Custom and Wingman.


dgibbs64:
Looks good, please ensure using tabs instead of spaces.

borzaka:
Done.

PR #2077, opened 2018-11-19
[WIP] New mods

dgibbs64:
This is a branch that I found. So I made a PR

### DIFF
--- a/lgsm/config-default/config-lgsm/csgoserver/_default.cfg
+++ b/lgsm/config-default/config-lgsm/csgoserver/_default.cfg
@@ -20,15 +20,15 @@
 # Wingman						0			2			
 # Danger Zone (FreeForAll)		6			0			mg_dz_blacksite
 gametype="0"
-gamemode="0"
+gamemode="1"
 mapgroup="mg_active"
 ip="0.0.0.0"
 port="27015"
 clientport="27005"
 sourcetvport="27020"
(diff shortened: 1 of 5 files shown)


Bara:
And [Accelerator](https://forums.alliedmods.net/showthread.php?t=277703) would be nice.

borzaka:
Many CSGO mods can't be used together. So basically besides Source mod (metamod, sourcemod, and some basic mod), only one mod installation is recommended, because probably they will break each other. Or if you are a pro in mods and server configuration, you can disable/enable mods, but after installation out-of-the box they won't work.

For example: Pug Setup and Get5 almost identical in function, but with different purpose. Installing both is not a good idea.

dgibbs64:
This will need further investigation

saintplay:
The user should be aware of the mods compatibility, if they install both mods, deleting the mods should be enough to fix the server.

This pull request needs some updates, there are some undocumented changes, code formatting, the urls don't match latest releases, etc.

I don't know if the author will continue to fix this issues.
I could create a new PR if it's necessary @dgibbs64

dgibbs64:
@saintplay I agree. These issues need to be addressed before merged. I won't be able to get to this for a while as its a lower priority. But If you are willing to contact the PR author and either collaborate or PR this to improve it that would be great. The mods commands do need some improvement.

PR #2099, opened 2018-12-07
Add CS:GO Danger Zone to Gamemode overview

Bara:


borzaka:
The mapgroup for the new gamemode is: `mg_dz_blacksite`
And probably only `dz_blacksite` map can be used with this gamemode.

PR #1751, opened 2017-12-27
Tmux session size arguments out of order.

Grimston:
Correct the width and height argument order for the tmux session.

PR #2103, opened 2018-12-10
fix(start): correct order of tmux size arguments

StrixG:


dgibbs64:
I had to revert this change previously but not sure why https://github.com/GameServerManagers/LinuxGSM/commit/69199a1af1bfcfdbc751ff217c040686956653ba

StrixG:
This change was reverted because it was merged into the wrong branch.
Your comment: https://github.com/GameServerManagers/LinuxGSM/pull/1751#issuecomment-354460215

PR #2110, opened 2018-12-13
fix(inssserver): fix the password and the default scenario parameters

Outpox:
Hey there,

I've had to fix this in my config by overwriting the `fn_parms()` function. The `password` field is a travel parameter according to [the doc](https://docs.google.com/document/d/1GDLg5p9jjeIya7EgBk0ibzDtDlyQ-U_jpspOzby-JmM/edit#heading=h.quyepjlz6tpc) and not a command line argument.  
I also fixed the default scenario which seems invalid as my server couldn't start. I've set it to one that I know works as I'm using this setting.  
  
I've followed your contributing guidelines regarding my commits.  
  
Cheers


dgibbs64:
Thanks for this :)

Outpox:
Awesome! :D

PR #2141, opened 2019-01-01
[WIP] Fix Insurgency Sandstorm password bug

### LINKED ISSUE #2137: Insurgency Sandstorm Server Password Bug
### LinuxGSM bug

Regarding server config in this directory:
~/lgsm/config-lgsm/inssserver/

When copying the contents of **_default.cfg** to **common.cfg** and entering a password in `serverpassword="123456"`, the game that appears in the server browser is not password enabled.

The bug lies in the server start command section of the same file. According to their [official Google Doc Sandstorm Server Admin Guide](https://docs.google.com/document/d/1GDLg5p9jjeIya7EgBk0ibzDtDlyQ-U_jpspOzby-JmM/edit), **password** is a Travel Parameter and not a Command Line Argument. Replace the preceding space+dash next to **password** with a question mark and the server password will be enabled when the server starts., but entering an empty password string into the variable makes the server passworded.

### Incorrect
`fn_parms(){
parms="${defaultmap}?Scenario=${defaultscenario}?MaxPlayers=${maxplayers}?Port=${port}?QueryPort=${queryport} -password=${serverpassword} -hostname='${servername}' -log"
}`

### Makes the server passworded regardless if the password variable is an empty string or not.
`
fn_parms(){
parms="${defaultmap}?Scenario=${defaultscenario}?MaxPlayers=${maxplayers}?Port=${port}?QueryPort=${queryport}?password=${serverpassword} -hostname='${servername}' -log"
}`

NYPD:
Closes #2137

### DIFF
--- a/lgsm/config-default/config-lgsm/inssserver/_default.cfg
+++ b/lgsm/config-default/config-lgsm/inssserver/_default.cfg
@@ -21,10 +21,10 @@ maxplayers="28"
 
 ## Server Start Command | https://docs.linuxgsm.com/configuration/start-parameters#additional-parameters
 fn_parms(){
-if [ "${serverpassword}" != "" ]; then
-parms="${defaultmap}?Scenario=${defaultscenario}?MaxPlayers=${maxplayers}?Port=${port}?QueryPort=${queryport}?password=${serverpassword} -hostname='${servername}' -log"
-else
+if [ -z "${serverpassword}" ]; then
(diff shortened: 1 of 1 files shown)


NYPD:
I can't get this this work on my Ubuntu server. It should be correct, but I'm not really experienced with shell scripts to know what is wrong. 

We cant have ?password to be set to an empty string because that will force the server to have a password anyway and prevent players from joining. The only way to have a server not have a password is to not include the travel parameter at all.

dgibbs64:
Thanks for this. I will hotfix this one later on

Issue #2109, opened 2018-12-12
7DTD Wrong query port Alpha 17


Eekiig:
**Environment**
OS: Ubuntu server 16.04.5 LTS
Server: 7DaysToDie Alpha 17

**Problem**
If you update the 7DaysToDie-Server to latest experimental (A17), the monitor tool restarts the server even if it is running.

**Reason**
In Alpha 17 the query-port has moved to the main gameserver-port but the monitor (gsquery) querys on gameserver-port+1.

**Workaround**
Change the queryport to the main port in info_config.sh
`nano ~/lgsm/functions/info_config.sh`
Edit queryport in section:
`fn_info_config_sdtd()`
from
`[row 909] queryport=$((port + 1))`
to
`[row 909] queryport=$(grep "ServerPort" "${servercfgfullpath}" | tr -cd '[:digit:]')`

Now the monitor will use the ServerPort for querys.

Greetings Eekiig

joshhsoj1902:
FYI, A17 was released today so this is now the version installed by default!

Eekiig:
Nope, doesn't work for me. Update to latest lgsm results in an empty queryport.
`Monitor sdtd-server: Querying port: gsquery: 192.168.0.21: : 0/1: QUERYING[ FAIL ] Monitor sdtd-server: Querying port: gsquery: 192.168.0.21: : 0/1: FAIL[ FAIL ]`

dgibbs64:
I did quite a bit of work on this to get everything working with A17 so should be working correctly now. How odd

Eekiig:
Seems the workaround on my first post didn't work anymore with the latest version of lgsm. It leads to the same result of an empty queryport.

dgibbs64:
I have re-tested this and querying is working for me. The query port is now the same port as the gameport which by default is 26900

```
DESCRIPTION  DIRECTION  PORT   PROTOCOL
> Game/RCON  INBOUND    26900  udp
> Query      INBOUND    26900  udp
> WebAdmin   INBOUND    8080   tcp
> Telnet     INBOUND    8081   tcp
```
I would check that the details are being pulled correct. The data is pulled from the config `sdtdserver.xml`

Eekiig:
After a reinstall of LGSM all works fine.
- Bug closed - Thanx! ;-)

PR #2142, opened 2019-01-03
fix(sdtdserver): update queryport for alpha 17

h3o66:
fix queryport for server monitoring the easy way


h3o66:
Related issue #2109

PR #2312, opened 2019-04-27
feat(backup): Update to allow use of backup exclusion file for omitting content from backups

jimmyatSplunk:
I've reworked the backup command to use a flat file that is used for file and folder exclusions in an attempt to allow admins to exclude files or folders from their backups. The idea being they can exclude game files that will be replaced by repairing the game server and it helps keep their backups smaller.

I've also added the necessary exclude file variable and a check for the "backupexclude.cfg" in the same fashion as "common.cfg" and "instance.cfg" so it should fold into current installations without issue.


jimmyatSplunk:
This request is essentially the same request as https://github.com/GameServerManagers/LinuxGSM/pull/1595#issue-137964814 but there are more changes in that that have nothing to do with implementing this improvement so it should be rejected and closed.

Frisasky:
please commit into develop branch as master branch are used for script release

dgibbs64:
Looks great :). I will need to review this to make sure I fully understand how this all works and do testing. So please bear with me on this might take a little time to merge

dgibbs64:
I think a backup re-factor is in order

linuxgsm.sh
```diff
@@ -20,7 +20,7 @@ if [ -f ".dev-debug" ]; then
 	set -x
 fi
 
-version="190301"
+version="190422"
```
h3o66: i guess this should not be in the commit, as the version is changed by release.

h3o66 (changes requested):
Please rebase and resolve the conflicts

PR #2353, opened 2019-05-12
fix(mtaserver): wrong executable name in update

DUDU54:
While looking for something, i saw that the name of the executable was not correct


dgibbs64:
@DUDU54 I need your github email so I can do a squash merge. Thanks

DUDU54:
@dgibbs64 
> 14166148+DUDU54@users.noreply.github.com

PR #2375, opened 2019-06-04
add get5 and SteamWorks mods

saintplay:
# Description

- Fix TTT and GOKZ mods not appearing in CS:GO mods
- Fix TTT and GOKZ removing sourcemod directories at mod deletion
- Add Get5 mod (CS:GO)
- Add SteamWorks mod (Sourcemod) 

Fixes #2077

## Type of change

* [x] Bug fix (change which fixes an issue).

## Checklist

## Provide Github Email

```
Co-authored-by: Diego Jara <diegojara6@gmail.com>
```


dgibbs64:
It there a risk of conflicting mods at all? as per this concern https://github.com/GameServerManagers/LinuxGSM/pull/2077#issuecomment-444657953

I did not develop much of the mod's functionality myself and dont know too much about mods.

saintplay:
General speaking: yes.

Removing mods and only re-installing the desired mod should fix any conflict.

> Many CSGO mods can't be used together. So basically besides Source mod (metamod, sourcemod, and some basic mod), only one mod installation is recommended, because probably they will break each other.

I think the CS:GO server admins are vey aware of this, and should't be a problem

borzaka:
In my opinion the current integrated SourceMod and Metamod:Sourcemod installation is a good and enough start, and if you want other custom mods, install and configure them manually for your needs. There are so many of them out there. That shouldn't be a problem. But if we let every to install TTT, GOKZ and Get5 all at once, it will be chaotic.

But I'am agree to add [Steamworks](https://partner.steamgames.com/), because that just an extension to functionality, and won't break any other mods. Some of them even require it.

saintplay:
@borzaka 

> That shouldn't be a problem. But if we let every to install TTT, GOKZ and Get5 all at once, it will be chaotic.

yes, but the workaround would be removing them, it should leave you with a clean CS:GO Server.
I think it's safe to say we can only guaranty stability for one CS:GO installation.

These conflicts already exists with TTT and GOKZ (and these even mess up sourcemod files when removing them).

There are a tons of mods in CS:GO, but I think we should only stick with well-maintained ones like Get5.

However a better solution would be warning a user or prompting again some message like:

```
You already have a CS:GO mod. Having multiples mods might cause conflicts and bugs. Do you want to continue? [Y/n]:
```

borzaka:
What if I want one TTT and one GOKZ server, and I'am using instances under the some user? So removing them and leaving only one is not an option. In this case I have to load/unload plugins on the instance level, but that is out of LinuxGSM's scope at the moment.

I know the workaround, you know the workaround, and many server operators know the workaround, but it's not known by common users who just whats to leverage the ease of use of the LinuxGSM.
Managing mods needs to be idiot proof in my opinion, and that's the point of LinuxGSM: it's working out-of-the box, and if you want to extend it, you can, but you need the knowledge. LinuxGSM can help you with the basics, but CS:GO mods are complicated to use together.

Other option would be to write mod managing script for only CS:GO mods which;
- only lets you install one mod besides the basic ones: SourceMod, Metamod, etc.
- lets you install another on a new instance
- helps you enable/disable and load/unload plugins on the instance level
- future proof, idiot proof, ease of use
- etc.

I agree, that we should stick with the well-maintained ones, but with only those, which can work together nicely, like Steamwork, SourceMod, Metamod, and probably some other extended functionality plugins.

And for example I prefer PugSetup over Get5 :P

saintplay:
> - only lets you install one mod besides the basic ones: SourceMod, Metamod, etc.
> - helps you enable/disable and load/unload plugins on the instance level
> - future proof, idiot proof, ease of use

Wouldn't be an additional prompt enough?

**You already have a CS:GO mod. Having multiples mods might cause conflicts and bugs. Do you want to continue? [Y/n]:**

borzaka:
Maybe, this is needs to be thoroughly think thru.

saintplay:
while enabling/disabling plugins for each instance would be awesome, it's outside of the scope of this PR.

what do you think of a second prompt for one-mod-only games like CS:GO @dgibbs64 ?

dgibbs64:
A prompt would be a good idea. I think a refactor of the mod's functionality is needed anyway in the future.

I have no idea about CS:GO mods.
Are some of these mods compatible with each other and won't cause a conflict?
If so Then they can be merged right away maybe in a new PR?
Any mods that are incompatible maybe should maybe be listed separately with a warning.

dgibbs64:
I have decided to accept this PR. However, a refactor of mods command is required in the future. Thank you :)

saintplay:
Thanks.

So a PR including the suggested prompt would make sense before or after that refactoring? @dgibbs64

PR #2376, opened 2019-06-04
[WIP] change csgo gamemode and tickrate

saintplay:
# Description

Change CS:GO defaults to competitive settings, since most people are going to use these

Fixes #2077

## Type of change

## Checklist

## Provide Github Email

```
Co-authored-by: Diego Jara <diegojara6@gmail.com>
```


borzaka:
Why `maxplayers="10"`? What about spectators, and GOTV bot?

If you want to limit how many user can join a team, use console commands, or a 3rd party mod for that, but do not limit the `maxplayers` to only support 5v5

saintplay:
What should be the default for `maxplayers`: 16, 20, 32, 64? Depends on the use case.

> but do not limit the maxplayers to only support 5v5

I don't think this limit anything, you can override the value in `_common.cfg` or `csgoserver.cfg`

I'm picking a value that in my opining should be the default (covers the most use cases)

borzaka:
Maxplayers 16 was a good default value.

What's the point setting a variable, and then override in somewhere else? Make a good one in the first place.

I forgot to mention coaches. If booth teams are having coaches, that's 12 players at least. Your maxplayer settings doesn't allow coaches, and any other spectators/bots to join the server. So that is not a good maxplayer value.

Anything higher than 10 is better. It could be 64, no one would care, and wouldn't make a difference, but lowering to 10 is limiting the game basic functionality.

image

You are editing a very old PR, since than many things changed, and many things can be outdated in this. And maybe there will be conflicts;
> This branch is out-of-date with the base branch

I would try to marge the latest changes from to develop, and it's gets messy, I would open a new PR for my individual changes.

saintplay:
questionable enough, changed to 16 players (again)

dgibbs64:
The only main concern with this PR is that it has the potential to mess up settings for any user that has left the default settings in place. So if an update was carried out it could change the game mode when rebooting the server. So this will have the be thought about

PR #2378, opened 2019-06-07
[WIP] Add support for FreeBSD

### LINKED ISSUE #2377: FreeBSD not supported
LinuxGSM does not work on FreeBSD out of the box. However, I have created a branch in which I did the minimum amount of porting to get it to run Minecraft on FreeBSD. I have opened this issue so I can link to it from the PR.

kyleedwardsny:
# Description

Add support for FreeBSD

Fixes #2377

## Type of change

## Checklist

# Distro Tests Required

* [ ] Ubuntu 16.04
* [ ] Debian 8
* [ ] CentOS 7
* [ ] Fedora 30

## Provide Github Email

Fill out below info or tick box below:
```
Co-authored-by: Kyle Edwards <kyleedwardsny@gmail.com>
```

- [ ] I do not wish to provide an email. I am aware this will hide me as the author of this commit.

All pull requests will now be squashed to create a tidy commit history and simplify changelog creation. You can provide either your own email or a GitHub-provided no-reply email.

When a PR is squashed the author becomes the person who squashed the PR. This removes you as the author of your own PR.
The only workaround for this is to add your details as a co-author. More info about co-authors can be found [here](https://help.github.com/en/articles/creating-a-commit-with-multiple-authors).


dgibbs64:
Tested this and installation does not work currently. Work needs to be completed on the installer and the dependencies check. This PR needs some more development. As I don't use FreeBSD I am limited on the amount I can help.

needed to install this to get started
```
pkg install bash wget 
```
had to change the bash location to this
```
#!/usr/local/bin/bash
```

Got this error
```
Installing Just Cause 2 Server
=================================
./steamcmd.sh: line 37: /usr/home/linuxgsm/jc2server/steamcmd/Steam.AppBundle/Steam/Contents/MacOS/steamcmd: No such file or directory
Warning! SteamCMD did not complete the download, retrying: Attempt 2
```

Frisasky:
well in order to run steamcmd under freebsd you need to install linux compatibility layer, even though both linux and freebsd superficially looks similar but their kernel and ABI are complete different
when i test steamcmd under freebsd 11 it did not work due to unsupported socket option, dunno how things going on freebsd 12

kyleedwardsny:
There's no ideal solution for the shebang, unfortunately. Changing it to `#!/usr/local/bin/bash` will of course break it for Linux users. I suppose we could change it to `#!/usr/bin/env bash`, or we could make the script patch its own shebang if it detects a FreeBSD installation.

FWIW, FreeBSD usually work around issues like this by just making a symlink: `/bin/bash -> /usr/local/bin/bash`. I've just been running `bash linuxgsm.sh`.

kyleedwardsny:
As far as the package installation... personally, I've never liked packages that automatically connect to my system's package manager without asking me. I would be OK with adding support for FreeBSD's `pkg` if there's a way to disable this behavior (some sort of "don't install any packages, let me install them myself" option.)

kyleedwardsny:
Also, FreeBSD has two different ways of installing software packages from their repository: Ports and `pkg`. `pkg` downloads binary packages similar to `apt-get`, `yum`, etc., while Ports download and build the source code (with the option to disable things that the user doesn't want/need.) We should not assume the user's preference in this regard.

kyleedwardsny:
As @Frisasky correctly pointed out, FreeBSD doesn't support some of the socket options that `steamcmd` uses. Granted, it might *happen* to work if you ignore the error, but this brings up a larger issue: most games aren't made for FreeBSD. Most of them certainly won't work without the Linux compatibility layer, and some won't work even with it. (Minecraft is a notable exception: it runs on Java, and works beautifully on the FreeBSD build of OpenJDK 8. The OpenComputers mod even has a built-in FreeBSD library for its Lua interpreter.)

To that end, I think we should either:

1. Blacklist games that are known not to work on FreeBSD, or
2. Preemptively only whitelist games like Minecraft that are known to work, and people can add support for others later on

kyleedwardsny:
@dgibbs64 What sort of workflow do you prefer? Do you want me to rebase and squash, or add commits on top of the merge commits you've already made?

dgibbs64:
Add commits on top I squash on merge.

lgsm/functions/check_permissions.sh
```diff
@@ -149,17 +163,11 @@ fn_sys_perm_errors_detect(){
 	sysdirpermerror="0"
 	classdirpermerror="0"
 	netdirpermerror="0"
-	# Check permissions/
-	# /sys, /sys/class and /sys/class/net should be readable & executable.
+	# Check permissions
+	# /sys, /sys/class and /sys/class/net should be readable & executable
 	if [ ! -r "/sys" ]||[ ! -x "/sys" ]; then
```
kyleedwardsny: FreeBSD does not have `/sys`. I had a check here for this. It looks like it got lost in the latest merge commit.

kyleedwardsny:
On closer inspection, it looks like one of the merges completely botched my code. I think I would rather rebase the commits onto the latest `develop` than try to duplicate/fix everything that got messed up.

dgibbs64:
I have never used rebase before. But I normally just squash a merge to allow one since commit once completed. What has gone wrong?

dgibbs64:
I agree with whitelisting only tested games. My other concern is I have no knowledge of freebsd so I am unable to properly support it. There is a risk of my development breaking a FreeBSD installation so FreeBSD would have to have a caveat of very limited support. There was a suggestion of using 
/usr/bin/env in the past but was rejected. Since distros have moved on since then it is possible that /usr/bin/env is perfectly fine to use. There will need to be tested on the oldest supported distros to ensure that there are no issues.
https://github.com/GameServerManagers/LinuxGSM/pull/1652

kyleedwardsny:
> What has gone wrong?

#2389 had a lot of merge conflicts with my code. Unfortunately, they didn't get resolved properly, which resulted in a lot of my code getting erased. In the end, I decided it was easier to rebase my commits on top of the latest `master` so I could resolve the conflicts properly. Everything appears to be back in order now.

> There was a suggestion of using
/usr/bin/env in the past but was rejected. Since distros have moved on since then it is possible that /usr/bin/env is perfectly fine to use.

Alternatively, if `/usr/bin/env` still creates an issue, we can leave it as it is and let the **extremely** small set of FreeBSD users (currently only me) either locally patch their script, add a `/bin/bash` symlink, or just run `bash linuxgsm.sh`.

> It will be great to support a new distro but we will have to tread carefully and I will need extra support to make it work. Good documentation for both user and developer will be a must for this.

I absolutely agree, and I completely understand that I have an extremely niche use case. Honestly, I would have been fine with just keeping my patches local, but I wanted to be a good open source citizen so that other potential FreeBSD users don't have to duplicate my work.

In short, I am totally fine with saying "we've made some meager attempts to support FreeBSD, but in the end, mainstream Linux is our priority, and you're on your own if something goes wrong."

lgsm/functions/check_deps.sh
```diff
@@ -261,6 +264,9 @@ fn_found_missing_deps(){
 			elif [ -n "$(command -v yum 2>/dev/null)" ]; then
 				cmd="sudo yum -y install ${array_deps_missing[@]}"
 				eval "${cmd}"
+			elif [ -n "$(command -v pkg 2>/dev/null)" ]; then
+				cmd="sudo pkg -y install ${array_deps_missing[@]}"
```
kyleedwardsny: I'm very confused as to why Codacy doesn't like this line... it is exactly the same as 3 lines above. Should I change this along with the existing lines to Codacy's suggestion?

kyleedwardsny:
I have confirmed via Docker that `/usr/bin/env` is present in all of the Linux distros you mentioned. I have not actually tested the script in them. Do you have a test suite I can run?

dgibbs64:
There has been no update on this for around 9 months. I am unable to support FreeBSD myself. I will be closing this as I do not think it is feasible to add support at this time.

PR #2401, opened 2019-06-17
Release v19.7.0

dgibbs64:


freaky-m0:
It's still the same problem. The script still fails because it looks for a UDP port which isn't there. 

./mcserver d gives:

```
Ports
=====================================================================================
Change ports by editing the parameters in:
/home/mcserver/serverfiles/server.properties

Useful port diagnostic command:
netstat -atunp | grep java

DESCRIPTION  DIRECTION  PORT   PROTOCOL
Game       INBOUND    25565  tcp
Query      INBOUND    25565  udp
Rcon       INBOUND    25575  tcp
```

But if I look with netstat it's just TCP:
```

root@gameserver:/home/user# sudo netstat -atunp | grep 25565
tcp6       0      0 11.22.33.44:25565       :::*                    LISTEN      16570/java
```

Issue #2183, opened 2019-02-06
Extra Parameters Option


dgibbs64:
## User Story

As a server admin, I want to add an extraparms option to remove the need to copy the whole parms function to add to a parameters config.

## Basic info

* LinuxGSM config behavior change

## Further Information

Now you can easily add any extra parameter to your gameserver without copying and modifying the whole fn_params() in your instance.cfg

PR #2434, opened 2019-07-09
Allow setting of level for, e.g. HapisIsland, or levelurl for custom maps

RFC1920:
This adds the ability to easily choose a custom level or custom map.


dgibbs64:
Looks good. I prefer to avoid too many if statements in the config though. I might move the if's to a module instead. It may also need to be sorted alongside the extra parms pr #2183

RFC1920:
FWIW the section I changed here was an if/else structure.  Agreed it should probably be somewhere else and also include the extraparms from #2183.

dgibbs64:
Need an opinion on this from a Rust admin

dgibbs64:
Features of this PR have been added via #2711 Thank you

Issue #2031, opened 2018-10-07
Add sv_lan and sv_cheats options to csgoserver default config file


borzaka:
The idea is came from the official [Counter-Strike: Global Offensive Dedicated Servers](https://developer.valvesoftware.com/wiki/Counter-Strike:_Global_Offensive_Dedicated_Servers) page.
Under [Advanced Configuration](https://developer.valvesoftware.com/wiki/Counter-Strike:_Global_Offensive_Dedicated_Servers#Advanced_Configuration) there is an example autoexec.cfg:

```
log on //This is set to turn on logging! Don't put this in your server.cfg
hostname "Counter-Strike: Global Offensive Dedicated Server"
rcon_password "yourrconpassword"
sv_password "" //Only set this if you intend to have a private server!
sv_cheats 0 //This should always be set, so you know it's not on
sv_lan 0 //This should always be set, so you know it's not on
exec banned_user.cfg
exec banned_ip.cfg
```

So, these two sounds like a great idea to add to csgoserver default config file:
> sv_cheats 0 //This should always be set, so you know it's not on
> sv_lan 0 //This should always be set, so you know it's not on

Benefits:
- you know it's not on, or it's on
- you could set this settings per instance: maybe you want an instance with `sv_cheats 1` for practice, and another one for only lan games (`sv_lan 1`), etc.

ohbus:
This issue has been addressed over [here in this merge](https://github.com/GameServerManagers/Game-Server-Configs/pull/68).

Please close this issue as it has now been **resolved.**

Thank you.

borzaka:
This issue is resolved via:
- [Game-Server-Configs #68](https://github.com/GameServerManagers/Game-Server-Configs/pull/68)
- https://github.com/GameServerManagers/Game-Server-Configs/commit/f5f339b59987b96985b9eeaa01b11f6f929abe2b#diff-e6740dbdb68468b3d1f7c1341bf1152f

I think the best practice is to set these values in the config files, not in the command line.

ohbus:
Kudos!

PR #2442, opened 2019-07-12
update default gamemode='1' and address the issue #2031

ohbus:
`commit` [0db1e2f](https://github.com/GameServerManagers/LinuxGSM/pull/2442/commits/0db1e2f07062c91dd54d2eb25db08d06b7a51351) deals with updating the default `gamemode='1'` to start the server as a **competitive match** rather than a casual match.

**and**

`commit` [0971a64](https://github.com/GameServerManagers/LinuxGSM/pull/2442/commits/0971a642345ea08a09da707405fba42714b1881e) deals with #2031 and servers to solve the issue with adding `sv_lan 0` and `sv_cheats 0` explicitly under the default starting parameters argument list.

### DIFF
--- a/lgsm/config-default/config-lgsm/csgoserver/_default.cfg
+++ b/lgsm/config-default/config-lgsm/csgoserver/_default.cfg
@@ -29,6 +29,8 @@ sourcetvport="27020"
 defaultmap="de_mirage"
 maxplayers="16"
 tickrate="64"
+svlan="0"
+svcheats="0"
 
 ## Required: Game Server Login Token
 # GSLT is required for running a public server.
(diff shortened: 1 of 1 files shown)


h3o66:
imho I would prefer only to take the second commit for the sv_lan changes and sv_cheats.
I guess it is not wanted to change all gameservers that use the current default for the servers that were already set up some time ago

ohbus:
> imho I would prefer only to take the second commit for the sv_lan changes and sv_cheats.
> I guess it is not wanted to change all gameservers that use the current default for the servers that were already set up some time ago

OK I understand your concern, then shall I remove the first commit?

ohbus:
> imho I would prefer only to take the second commit for the sv_lan changes and sv_cheats.
> I guess it is not wanted to change all gameservers that use the current default for the servers that were already set up some time ago

When i had setup the server with my friends All of them suggested that the default game mode should be changed to competitive as that is the most widely played game mode and most preferred way of playing even with custom maps like 5v5 Retake Servers.

i hope you understand the concern and merge this commit for the good of the counter strike community where I believe this should have been default playing mode in the game as done by the default vanilla in-game screen as well.

borzaka:
`sv_lan` and `sv_cheats` are already in the csgoserver game server configs:
https://github.com/GameServerManagers/Game-Server-Configs/blob/master/CounterStrikeGlobalOffensive/server.cfg

No need to set it via command line as well. Option to set these in two places is a bad practice. They can override each other. As in the #2031, even in the official documentation they setup these values in the config file, not in the command line.

I agree with the default game mode set to competitive.

ohbus:
> `sv_lan` and `sv_cheats` are already in the csgoserver game server configs:
> https://github.com/GameServerManagers/Game-Server-Configs/blob/master/CounterStrikeGlobalOffensive/server.cfg
> 
> No need to set it via command line as well. Option to set these in two places is a bad practice. They can override each other. As in the #2031, even in the official documentation they setup these values in the config file, not in the command line.
> 
> I agree with the default game mode set to competitive.
> 
> >  GitHub**[GameServerManagers/Game-Server-Configs](https://github.com/GameServerManagers/Game-Server-Configs)**A repo of game server configuration files. Contribute to GameServerManagers/Game-Server-Configs development by creating an account on GitHub.

Okay!

I did not notice that those config files are used when the server got deployed.

Assumed it to be some side project.

Thank You @borzaka for bringing up.

PR #2502, opened 2019-09-09
hotfix v19.10.1

dgibbs64:


sasjafor:
I forked and merged this PR on the fork and tested it on Linux. It does download oxide now, but it doesn't seem to apply it correctly. If I ran mods-update in succession, it would always download again and after manually updating oxide, the command still downloads oxide.

PR #2822, opened 2020-04-13
feat(steamcmd): move steamcmd dir to standard location

dgibbs64:
# Description

After discussion with a Debian maintainer @smcv, some research and testing I have confirmed the standard steamcmd directory. This feature standardises the steamcmd directory with both the apt package and the curl download. 

Depending upon which version of the apt package or curl download steamcmd installed, it will use one of several directories.
```
/home/lgsm/.steam
/home/lgsm/Steam
/home/lgsm/.local/share/Steam
/home/lgsm/steamcmd
```
Using a workaround highlighted https://github.com/ValveSoftware/steam-for-linux/issues/6976#issuecomment-610446347 by @smcv I have updated the steamcmd installer so SteamCMD will always be the correct latest install directories.

Fixes #[issue]

## Type of change

* [X] New feature (change which adds functionality).

## Checklist

* [ ] This pull request Subject follows the Conventional Commits standard.
* [ ] This code follows the style guidelines of this project.
* [ ] I have performed a self-review of my own code.
* [ ] I have checked that this code is commented where required.
* [ ] I have provided a detailed enough description of this PR.
* [ ] I have checked If documentation needs updating.

## Documentation


lgsm/functions/check_steamcmd.sh
```diff
@@ -63,6 +63,43 @@ fn_check_steamcmd(){
 	fi
 }
 
+fn_check_steamcmd_dir(){
+	# Worksround that pre-installs the correct steam directories to ensure all packages use the correct Standard
+	# https://github.com/ValveSoftware/steam-for-linux/issues/6976#issuecomment-610446347
+	if [ ! -d "${HOME}/.local/share/Steam" ]; then
+		mkdir -p "${HOME}/.local/share/Steam"
```
smcv, line 70: Please use `${XDG_DATA_HOME}/Steam` if set, as per the XDG basedir spec: https://specifications.freedesktop.org/basedir-spec/latest/

The easiest way to do this is probably to do:

```
: "${XDG_DATA_HOME:="${HOME}/.local/share"}"
```

and then use `${XDG_DATA_HOME}/Steam` everywhere else.

lgsm/functions/check_steamcmd.sh
```diff
@@ -63,6 +63,43 @@ fn_check_steamcmd(){
...
+			rm "${HOME}/.steam/root"
+		fi
+		ln -s "${XDG_DATA_HOME}/Steam" "${HOME}/.steam/root"
+	fi
+
+	if [ ! -L "${HOME}/.steam/steam" ]; then
+		if [ -d "${HOME}/.steam/steam" ]; then
+			rm "${HOME}/.steam/steam"
+		fi
+		ln -s "${XDG_DATA_HOME}/Steam" "${HOME}/.steam/steam"
```
dgibbs64, line 88: @smcv Just wanted to check this is what you mean.

smcv: When I mentioned `~/.local/share/Steam` on the Steam bug, I was using it as shorthand for what the XDG Base Directory specification says:

* if `XDG_DATA_HOME` is set to a non-empty value, use`${XDG_DATA_HOME}/Steam` as the default for a new Steam installation directory
* if `XDG_DATA_HOME` isn't set or is set to an empty value, use `~/.local/share/Steam` as the default for a new Steam installation directory

The canonical reference for where Steam gets installed is `bin_steam.sh` in a full Steam installation. You can get a relatively up-to-date copy from the Steam launcher/bootstrapper, which is available from https://repo.steampowered.com/steam/archive/precise/ (download and unpack `steam_latest-beta.tar.gz` or `steam_latest.tar.gz` - currently the latest beta is 1.0.0.62 and the latest stable release is 1.0.0.61).

There are two common ways to implement `XDG_DATA_HOME` correctly in shell script:

1. Whenever you want to know the XDG base directory for data, use `${XDG_DATA_HOME:-"${HOME}/.local/share"}`. The `:-` means: if the variable `XDG_DATA_HOME` is set to a non-empty value, "return" that value. Otherwise, "return" the result of expanding the `HOME` variable in `"${HOME}/.local/share"`. This is how `bin_steam.sh` does it. Or:

2. Near the beginning of your script, use `${XDG_DATA_HOME:="${HOME}/.local/share"}` (typically as a parameter to a `:` or `true` so that it doesn't do anything). The `:=` means: if the variable `XDG_DATA_HOME` is set to a non-empty value, "return" that value. Otherwise, *set the variable* to the result of expanding the `HOME` variable in `"${HOME}/.local/share"`, and then "return" the new value. After doing that, you can use `${XDG_DATA_HOME}` unconditionally, because its value has already been set to a default if necessary.

It looks as though in this MR, you're doing a mixture of (1) and (2) by using `${XDG_DATA_HOME:="${HOME}/.local/share"}` everywhere, which is slightly excessive but should also work fine.

dgibbs64:
https://github.com/ValveSoftware/steam-for-linux/issues/6976

Issue #2117, opened 2018-12-15
[Backup] If server saves point to ${HOME} they may not be backed up


dgibbs64:
As a server admin if my game server is not saved in ${HOME} then some files may not be backed up.

**Describe the bug**
The issue arrises if the game server is not saved in ${HOME}. Backups are taken from ${rootdir} which is the top level directory of linuxgsm where ./gameserver is located. Some game servers choose to place some files in ${HOME}. Should an admin have there server in `/home/lgsm/sdtdserver` for example `/home/lgsm/.local/share/7DaysToDie` does not get backed up.

**To Reproduce**
have the game server in a non-standard directory

**Expected behaviour**
All game server files should be backed up.

**Additional context**
It is recommended that a server admin uses the HOME directory currently however if non-standard directories are used some check that you have a backup of all files.

PR #2867, opened 2020-05-05
Fix for Bug #2117

erdoking:
"#2117 [Backup] If server saves point to ${HOME} they may not be backed up"

This fix makes the backup include the savegames also on non standard installations.
Checked with sdtd.


dgibbs64:
Different game servers may have different save locations within HOME so this fix may not work with all game servers. I will take a look at this when I do planned backup refactor. I will hold this fix until then.

Thanks for creating this PR

Issue #2920, opened 2020-06-09
Unnecessary `sessionname` constraints


j4s0n:
## User Story

1. sessionname is derived from an Instance name. Many supported games can be run as clusters, where multiple Instances may share a name, but differ in other ways (ie, maps). However, the tmux sessions utilize only the sessionname, which adds an unnecessary uniqueness constraint to session names. 

Suggestion: This can be eliminated by appending the server port to the tmux sessions name..

2. The `grep` call in `check_status.sh` embeds the $sessionname inside a regular expression without escaping its contents. This adds an unnecessary constraint to sessionname contents: `sessionname may not contain any regular expression special characters`.

Suggestion: use bash sub-string matching instead of a grep regex

## Basic info

* **Distro:** Docker
* **Game:** Various
* **Command:** start
* **LinuxGSM version:** [v20.1.3]

## To Reproduce

1. Create 2 instances with the same sessionname. Start the first -- no problem. Start the second and it will not start because check_status.sh matches the first session and thus thinks the requested instance is already running.
2. Name your server 'Something [not a regex]'

## Expected behavior

Neither constraints are necessary, and can easily be removed. I'll submit a pull request demonstrating how I achieved this.

dgibbs64:
Been playing with the PR but had issues. Just to be clear here is an example of the issue you are trying to fix

Multiple installs with the same instance name will have the same session name so will conflict
/home/arkserver/arkserver <---- will conflict
/home/arkserver/arkserver-2
/home/arkserver-2/arkserver <---- will conflict

h3o66:
Done in Release 20.6.0

PR #2921, opened 2020-06-09
Fixed uniqueness constraints on session name

j4s0n:
See issue #2920


dgibbs64:
I can no longer merge this. I will hold just tin case you are doing hacktoberfest and make sure you the co-auther of the final pr

PR #3059, opened 2020-09-28
fix(mcserver): Support installing previous minecraft versions via branch

joshhsoj1902:
# Description

When the branch for minecraft is set, This should allow lgsm to download a previous version. 

I haven't actually tested this in LGSM yet, I only tested the updated jq command manually:
```
curl -s "https://launchermeta.mojang.com/mc/game/version_manifest.json" | jq -r '.versions | .[] | select(.id=="1.15") | .url'
```

Previously it would always grab the latest MC version, even if the branch was set. 

The branch [defaults](https://github.com/GameServerManagers/LinuxGSM/blob/master/lgsm/config-default/config-lgsm/mcserver/_default.cfg#L20) to "release", so as long as no one has changed that this shouldn't impact anyone's current configs 

Fixes #[issue]

## Type of change

* [ ] Bug fix (a change which fixes an issue).

## Checklist

## Documentation

### DIFF
--- a/lgsm/functions/update_minecraft.sh
+++ b/lgsm/functions/update_minecraft.sh
@@ -10,7 +10,7 @@ fn_update_minecraft_dl(){
 	if [ "${branch}" == "release" ]; then
 		latestmcreleaselink=$(curl -s "https://launchermeta.${remotelocation}/mc/game/version_manifest.json" | jq -r '.latest.release as $latest | .versions[] | select(.id == $latest) | .url')
 	else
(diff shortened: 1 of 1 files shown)


joshhsoj1902:
The one thing I haven't covered in this is what happens if a branch doesn't exist. 

I think the logic in [here](https://github.com/GameServerManagers/LinuxGSM/pull/3059/files#diff-44939efe877a1a32a59ca58e19cb73e3R86) will kick out if the jq command doesn't find anything. but it would also be possible to throw a better error if that we preferred.

dgibbs64:
My review of Minecraft is coming up so this is good timing

dgibbs64 (commented):
Happy to merge this once confirmed tested and working.

PR #3073, opened 2020-10-08
feat: allow override of repo with master

### LINKED ISSUE #3077: allow override of repo with master
When using an external (forked) repo, the master branch of that repo becomes useless. Reason is that the scripts have a hardcoded override that will change behaviour when using the master branch. Thus using the original repo instead of the forked one.

Some prefer using a forked repo to work with, so they have control over when the LGSM scripts are updated. For production type applications this is a basic necessity in my opinion.

mvaessen:
# Description
When using an external (forked) repo, the `master` branch of that repo becomes useless. Reason is that the scripts have a hardcoded override that will change behaviour when using the `master` branch. Thus using the original repo instead of the forked one.

Some prefer using a forked repo to work with, so they have control over when the LGSM scripts are updated. For production type applications this is a basic necessity in my opinion.

Fixes #3077
Added some extra conditions so the exception only occurs when the original repo is configured within the scripts.

## Type of change

* [X] Bug fix (a change which fixes an issue).

## Checklist

* [X] This pull request uses the `develop` branch as its base.
* [ ] I have performed a self-review of my code.
* [ ] I have checked If documentation needs updating.

## Documentation


README.md
```diff
@@ -3,7 +3,7 @@
...
...mz0ub%2FfuxEacBg%2FDmYtiCjgo5NG2mBXq%2BH5I1ogMRk9Zbd%2BQU2e1ML6VPLOyf5tvBQ8yT1lG10imxsABm7SLs898GTpyYynEzP60hO3trHDKvMigUwdeaceacqzp7nOI4n0SSIIjl36ao4Z356OV07fSQAk6xJ3XGg%2BLCr1d1OYlVHp4eUHPnerU79ZA%2F1kuv1JQMAg%2BE4O2P23EumF3VkvHprsZKMzKwbRUXFEyTvSIEmTVbrysp%2BWr8wfQHGK6WChVa3bKUmdWou%2BjpArdGkzZ41c1zG%2Fu5uGH4swzd561F%2BuhIT4%2BLnSuPsv9%2BJKIpjNr9dXYOyk7%2FBZrcjIT4eCnoKgedJP4BEqhG77E3NKP31FO7cfQA5K0dSYuLgz2TwCWJSOBzG6crzKK%2BohNfni%2Bx6OMUMMNe%2Fgf7ocbw0v0acKg6J8Ql0q%2BT%2FAXR5PNi5dz9c71upuQqCKFAD%2BYhrZLEAmpodaHO3Qy6TI3NhBpbrshGtOWKOSMYwYGQM8nJzoFJNxP2HjyIQho4PewK6hBktoDcUwtIln4PjOWzflQ%2Be5yl0yCCYgYikTclGlxadio%2BBQCSiW1UXoVGrKYwH4RgMrjU1HAB4vR6LzWYfFUCKxfS8Ftk5qxHoCUQAUkRJaSEokkV6Y%2F%2BJUOC4hn6A39NVXVBYeNP8piH6HeA4fPbpdBQV5KOx0QaL1YppX3Jgk0TwH2Vg6S3u%2BdB91%2B%2FpuNYPYFl5uP5V7ZqvsrX7jxqMXR6ff3gCQSTzFI0a1TX3wIs8ul%2Bq4HuWAAiM39vhOuR1O1fQ2gT%2F26Z8Z5vrl2OHi9OXZn995nLV9aFfS6UC9JeJPfuK0NBohWpCHMSAAsFe74WWP%2BvT25wtP9Bpob6uGqqyDnOtaeumjRu%2ByFu36VntK%2FPA5umTJeUtPWZSU9BCgud661odVp3DZtkc7AnYR33RRC708PrVi1larW7XwZIjLnd7R6SgSqWSNjU1B3F72pz5TZbXmX5vV81Yb7Lg7XT%2FUXriu8XLVqw6c6XqWnBKiiYU%2BMt3wWF7u7i91XlSEITwSAZ%2FCzAAHsJVbwXYFFEAAAAASUVORK5CYII%3D)](https://www.gnu.org/software/bash/)
```
h3o66, line 6: Why is this changed ?
This should be not included

tests/tests_fctrserver.sh
```diff
@@ -167,7 +167,7 @@ fn_bootstrap_fetch_file(){
 fn_bootstrap_fetch_file_github(){
 	github_file_url_dir="${1}"
 	github_file_url_name="${2}"
-	if [ "${githubbranch}" == "master" ]&&[ "${commandname}" != "UPDATE-LGSM" ]; then
```
h3o66, line 170: Why did you modify the tests ?
Are these as well needed for this change ?

h3o66 (changes requested):

README.md
dgibbs64: might be that I just updated this badge

tests/tests_fctrserver.sh
dgibbs64: the tests should partly match linuxgsm.sh

README.md
mvaessen: looking at the branches, it seems you only pushed to master, not develop.

tests/tests_fctrserver.sh
mvaessen: i did a replace all. Would suggest having consistency throughout, to prevent random behaviour within the code

dgibbs64 (approved):
So the standard behaviour for master is that it uses the release/version number rather than the latest master. This is designed to prevent modules "version mixing" e.g alerts.sh being on v20.4.0 and alerts_discord.sh being v20.5.0. Version mixing can cause unpredictable behaviour and break installs.

However, this behaviour causes a problem for forked repos using master as a fork may contain custom code on top of a release. It means that the custom code will be ignored by LinuxGSM.

Your solution will ensure that only the official GameServerManagers repo uses this method which is a good solution. It does mean there is the potential for unpredictable behaviour if the forked repo is updated but the installation is not and it then downloads an updated file from the updated fork causing "version mixing". To combat this it would be up to the fork maintainer to make sure there linuxgsm install regularly checks for updates to keep then synced with the fork.

The other solution would be for a fork to create their own version however most people would not know to do this and add complications, so I don't think that would work. So I think your solution is probably the best one to go with. Unless any other better solutions are suggested

dgibbs64:
Also please could you create an issue for this PR and link it. Thanks

dgibbs64:
I think it is also worth popping a comment on this if statement so it is understood what it is doing. @mvaessen fee frlee to add a comment. If you are struggling I can edit or add a comment

mvaessen:
@dgibbs64 feel free to take over from here and change it to what works best for you :)

PR #3272, opened 2021-02-18
fix(vhserver): Add escaped quotes to Valheim startparams to account for spaces

Log1x:
# Description

The current single tick quotes do not work as intended when using spaces in the server name. 

I instead used escaped quotes (tested + working) and added it to the password and world name for sanities sake.

## Type of change

* [x] Bug fix (a change which fixes an issue).

## Checklist

### DIFF
--- a/lgsm/config-default/config-lgsm/vhserver/_default.cfg
+++ b/lgsm/config-default/config-lgsm/vhserver/_default.cfg
@@ -17,7 +17,7 @@ gameworld="${selfname}"
 public="1"
 
 ## Server Parameters | https://docs.linuxgsm.com/configuration/start-parameters#additional-parameters
-startparameters="-name '${servername}' -password ${serverpassword} -port ${port} -world ${gameworld} -public ${public}"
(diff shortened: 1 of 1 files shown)


dgibbs64:
Thanks for this. I want to confirm the standards for quoting across LinuxGSM. I am not a fan of the `\"` because I think it can confuse users. However we must of course do what works. This will be considered for a hotfix

Log1x:
> I am not a fan of the \" because I think it can confuse users

Definitely agree. The main thing is using a single `'` in the config file would then require apostrophes to be escaped – it might come down to choosing the lesser evil unless some magic is done behind the scenes.

I've noticed this game in particular seems to be finicky when passing arguments to the executable FWIW.

dgibbs64 (approved):
yeah, there might need to be some behind the scenes jiggery-pokery to not use `\"` however at least for now this fix will do the job just fine :)

dgibbs64:
@Log1x could you please raise an issue and link it to this PR. Ready to merge once done. Thanks for your help 🚀

PR #3773, opened 2022-02-03
Only compare first 5 chars of distroversion

dhensby:
# Description

I've been trying to work on a simple cloud-init script for provisioning a cloud server with LGSM / csgoserver. I noticed in the output of the install the message `Warning! LinuxGSM dependency checking currently unavailable for Ubuntu 20.04.3 LTS.` was output and the required dependencies were not installed. Note the version number is `20.04.3` not `20.04`.

This fix should mean that only the first 5 characters of the versions number are compared (eg: 20.04 instead of 20.04.3).

Fixes #[issue]

## Type of change

* [x] Bug fix (a change which fixes an issue).

## Checklist

## Documentation

### DIFF
--- a/lgsm/functions/check_deps.sh
+++ b/lgsm/functions/check_deps.sh
@@ -19,11 +19,11 @@ fn_install_mono_repo(){
 		sleep 1
 		echo -en "   \r"
 		if [ "${distroid}" == "ubuntu" ]; then
-			if [ "${distroversion}" == "20.04" ]; then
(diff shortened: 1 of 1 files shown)


dhensby:
I've tried deploying this fix and it doesn't seem to be the problem at hand.

I'll do some more digging, but it looks like if you run the installer as root first, the CSV is not downloaded and so fails, then you run as non-privileged user and it does download it.

PR #3745, opened 2022-01-12
feat(core): move default settings to seperate conf

h3o66:
# Description

Move the notification part to a seperate config and prevent users from making changes there.

Fixes #3741

## Type of change

* [ ] Bug fix (a change which fixes an issue).
* [x] New feature (change which adds functionality).
* [ ] New Server (new server added).
* [ ] Refactor (restructures existing code).
* [ ] Comment update (typo, spelling, explanation, examples, etc).

## Checklist

PR will not be merged until all steps are complete.

* [ ] This pull request links to an issue.
* [ ] This pull request uses the `develop` branch as its base.
* [ ] This pull request Subject follows the Conventional Commits standard.
* [ ] This code follows the style guidelines of this project.
* [ ] I have performed a self-review of my code.
* [ ] I have checked that this code is commented where required.
* [ ] I have provided a detailed with enough description of this PR.
* [ ] I have checked If documentation needs updating.

## Documentation

If documentation does need updating either update it by creating a PR (preferred) or request a documentation update.
* User docs: https://github.com/GameServerManagers/LinuxGSM-Docs
* Dev docs: https://github.com/GameServerManagers/LinuxGSM-Dev-Docs

**Thank you for your Pull Request!**

PR #3795, opened 2022-02-17
feat(core): prefix logs with a timestamp

### LINKED ISSUE #3778: [FEATURE] Add timestamps to logs that don't support them
## User Story

As a server admin, I want to be able to view timestamps for my server logs.

## Basic info

* **Distro:** [Debian 11]
* **Game:** [Test on Rust, but probably applicable to others]
* **Command:** [Start]
* **LinuxGSM version:** [latest]

## Implementation

I have tried to implement this after interacting with the Github mods gentlemen on Discord.

I have found that we can prepend all lines of log with the timestamp by modifying these lines:

- https://github.com/GameServerManagers/LinuxGSM/blob/fdfeae1fcdccb89f55e41ca33e964bbcb035d2ce/lgsm/functions/command_start.sh#L95
- https://github.com/GameServerManagers/LinuxGSM/blob/fdfeae1fcdccb89f55e41ca33e964bbcb035d2ce/lgsm/functions/command_start.sh#L111

Should be changed to:

```bash
tmux pipe-pane -o -t "${sessionname}" "exec bash -c \"cat | $add_ts\" >> '${consolelog}'"
```
With a global variable (using `gawk` for readability)
```bash
add_ts='gawk '"'"'{ print strftime(\"[%Y-%m-%d %H:%M:%S]\"), \$0 }'"'"''
```

Or without gawk we can just put this script in a one liner:

```bash
while IFS= read -r line
do 
    printf "[%s] %s\n" "$(date '+%Y-%m-%d %H:%M:%S')" "$line";
done
```

I tested this on my rustserver, and the output:

```
Asset Warmup (1/9092)
Asset Warmup (408/9092)
Asset Warmup (434/9092)
Asset Warmup (1088/9092)
```

Becomes:

```
[2022-02-08 18:03:26] Asset Warmup (1/9092)
[2022-02-08 18:03:26] Asset Warmup (408/9092)
[2022-02-08 18:03:26] Asset Warmup (434/9092)
[2022-02-08 18:03:26] Asset Warmup (1088/9092)
```

I didn't add a check to manage lines with whitespaces because that would make the code too complex and unreadable.

We can also use `ts` from `moreutils` package but it adds dependencies:

```
tmux pipe-pane -o -t "${sessionname}" "exec bash -c \"cat | ts '[%Y-%m-%d %H:%M:%S]' \" >> '${consolelog}'"

```

We can also add a global variable to activate timestamps on certain games, not all. And a variable for the format.

Although this solution is not ideal, because the timestamp won't be precise for every line, but each time tmux dumps logs to the file, every line included in that dump will have the same timestamp of the moment the dump occurs. It still is much better than no timestamps at all in my opinion.

I'd like your feedback on this. Would you like something like this to be included in your code? Do you think there is something else I should add for this to be included in your codebase?
I'd like your feedback on this.

DrissiReda:
# Description

Adds the possibility to prefix each line of log with a timestamp before pushing it to the log file.

Currently I only enabled it on rustserver, because that's what I used to test it. But it should work for everything else. 

I pushed the modification to every file, but if #3745 is merged, I'll refactor my PR.

Fixes #3778 

## Type of change

* [ ] Bug fix (a change which fixes an issue).

## Checklist


h3o66 (changes requested):
the standard on lgsm is more "on/off" and not "True/False"
Please refactor this one first

dgibbs64:
Great idea!!

DrissiReda:
> Great idea!!

You think it needs anything else to be merged?

lgsm/functions/command_start.sh
```diff
@@ -8,6 +8,7 @@
 commandname="START"
 commandaction="Starting"
 functionselfname="$(basename "$(readlink -f "${BASH_SOURCE[0]}")")"
+add_ts="gawk '{ print strftime(\\\"[$logtsformat]\\\"), \\\$0 }'"
```
jusito, line 11: Suggesting to use date instead of gawk:
`printf '[%s] %s\n' "$(date "+$logtsformat")" "hello world"`
or `echo "asd" | xargs printf '[%s] %s\n' "$(date "+$format")"`
provides the same output as
echo "hello world" | gawk '{ print strftime("[%Y-%m-%d %H:%M:%S]"), $0}'
and we wouldnt need an additional dependency

DrissiReda: I had a hard time escaping that command to be able to use it in my exec . If you managed to do it I'd be happy to use it.

jusito: Simple explanation:
bash knows mostly strings, 
`"interpret special characters >once< before passing this to something"`
`'dont interpret special characters before passing'`

Therefore:
`echo "$UID" '$UID' # whitespace in between = two strings = two args for echo`  1st is checked for special characters and after that added as arg to echo, 2nd not
If you dont want this behavior use single quotes.

If you need to mix it, you can combine it.
`echo "$UID"'$UID' # no whitespace in between = one string, first part is interpreted before`
or you can escape with \
`echo "\$UID" '$UID' # output is identical`
if you do want to output \ between "", you need to escape it as well:
`echo "\\" '\' # but not if you use single quotes`
What if you want to use double or single quotes?
`echo "\"" '"'` or `echo "'"`
You can also escape whitespaces, so this is one string
`foo=hello\ world` but please dont do so

In your command: `add_ts="gawk '{ print strftime(\\\"[$logtsformat]\\\"), \\\$0 }'"` `add_ts=".."..".."` you used the "interpret special characters way" but you wanted to add double quotes to it, so you needed to escape it once. `add_ts="..\"..\".."` but why are the additional \ needed?
```bash
foo=hello
bar=\$foo
echo "$bar" # prints $foo because interpreted once, even if special symbol is in
eval echo "$bar" # prints hello but why?
# calls eval "echo" '$foo'
# eval interpreted once more and executes "echo" "hello"
#
# if you want to prevent it ( and how you did )
bar=\\\$foo
eval echo "$bar" # prints now $foo instead of hello
# thats what happened to you
```

Sry was more motivated to clarify this for you than fixing the issue.

DrissiReda: I understand all that, but I need it to process the variable `$logtsformat` inside of `add_ts`. Which means I have:


```bash
add_ts=gawk '{ print strftime(\"[%Y-%m-%d %H:%M:%S]\"), \$0 }'
``` 

Then in L96 the line will become:

```bash
tmux pipe-pane -o -t "${sessionname}" "exec bash -c \"cat | gawk '{ print strftime(\"[%Y-%m-%d %H:%M:%S]\"), \$0 }'\" >> '${consolelog}'"
```

I can put everything manually in that line, but we'll have the some logic repeated in multiple steps and it'll make the code less readable and harder to maintain.

dgibbs64 (approved):

PR #3895, opened 2022-06-16
Networking and server rebuild update

Neouni:
Networking and the server has been rebuild from scratch many commands have been lost and shutdown has changed.

# Description
Removed:
ip, port, queryport, autosaveinterval, clearinterval
These are now all defined in setting.xml and can't be defined on the commandline anymore
default.ini got removed

Modified:
worldtype & worldname to correct commands
stopmode changed, it needs a quit command now in the console

# Comment
The network changes have been live now for 2 weeks, server has been updated for 2 days expect more changes to follow.


h3o66 (changes requested):
Please rebase this onto the develop branch

PR #3836, opened 2022-04-03
fix(monitor): provide always valid exitcode

jusito:
# Description

[These changes are the core changes needed for my docker pull request.](https://github.com/GameServerManagers/LinuxGSM-Docker/pull/43) The changes are tested against every servercode.

1. command monitor had multiple issues, mainly it doesn't exit with a usable exit code. I applied the style guide and documentation should be correct now.
2. 127.0.0.1 should be part of fallback ips and maybe the others should even be removed. E.g. the port config inside a container can be 27015 but its mapped to host 27016, maybe because on 27015 is another server. Therefore if the host ip is taken to do a check, the wrong server is monitored.
3. `lgsm/functions/core_dl.sh:368` hides curl failing
4. `linuxgsm.sh` changes are really useful to debug multiple invocations. As a result you can run a clean container and dump this file after the run to get a complete debug log.
5. `lgsm/functions/install_server_files.sh` missed a check if branch is missing
6. `lgsm/functions/fix.sh` refactoring to get rid of lots of branches which are error prone. Also added check to validate names added to arrays are valid.
7. `lgsm/functions/install_header.sh` probably a typo?

Fixes #[issue]

## Type of change

* [x] Bug fix (a change which fixes an issue).
* [ ] New feature (change which adds functionality).
* [ ] New Server (new server added).
* [x] Refactor (restructures existing code).
* [ ] Comment update (typo, spelling, explanation, examples, etc).

## Checklist

PR will not be merged until all steps are complete.

* [x] This pull request links to an ~issue~ draft pr which depends on this.
* [x] This pull request uses the `develop` branch as its base.
* [ ] This pull request Subject follows the Conventional Commits standard.
* [x] This code follows the style guidelines of this project.
* [x] I have performed a self-review of my code.
* [x] I have checked that this code is commented where required.
* [x] I have provided a detailed with enough description of this PR.
* [x] I have checked If documentation needs updating.

## Documentation

If documentation does need updating either update it by creating a PR (preferred) or request a documentation update.
* User docs: https://github.com/GameServerManagers/LinuxGSM-Docs
* Dev docs: https://github.com/GameServerManagers/LinuxGSM-Dev-Docs

**Thank you for your Pull Request!**

PR #4155, opened 2023-04-01
fix(monitor): provide valid exitcode

jusito:
break down of: https://github.com/GameServerManagers/LinuxGSM/pull/3836

Only the part where you would like to see logging to console and when for fn_script_ , I would need a hint for that.


jusito:
regarding loging, read your comments on: #4156 
I think this helped with my confusion of fn_script_log_* vs fn_print_*
In my testing environment I always have a log which contains only fn_print_* or the debug log. But basically fn_script_log is the overall logfile for lgsm.
1. Do you want to add the fn_script_log_* , should I add them or do you want to not modify logging in this pr but rather later?

With this specific pr you can easily find issues in current servercodes, some of them I traced some time ago, many just dont open the queryport even if config is correct or have incorrect port config, or have missing queryport set and so on.
2. Are you interested in fixing these in v23.2.0 or rather v23.2.1 or above?
I can look into it, I have a complete log of every servercode with this change applied, so I could start tracing it. But I would probably need a week for it.

PR #4186, opened 2023-04-11
Feature/Steam Workshop support

FliesWithWind:
# Description

For a while, I had a horrible but working script for downloading and updating Arma 3 mods from Steam Workshop. I figured it was about time I try adding it to LinuxGSM. The script was based on https://github.com/arkmanager/ark-server-tools

Changes in this PR are working, but are far from being pretty or complete in my opinion. I'm no bash magician, so I wanted to show it as it is, hoping for an initial review and suggestions.

I've added it as a separate module, to avoid conflicts with the current mods module.
For now, it downloads mods for Arma 3 taken from the server config file. 
```
workshopmods="450814997;2131302796"
```

Once we reach a working and pretty solution, the next steps would be to add support for other games, like Arc and Starboud, and add Steam Collections support.

Fixes #[3514] #[1894] #[1623] #[960]

## Type of change

-   [ ] Bug fix (a change which fixes an issue).
-   [x] New feature (a change which adds functionality).
-   [ ] New Server (new server added).
-   [ ] Refactor (restructures existing code).
-   [ ] Comment update (typo, spelling, explanation, examples, etc).

## Checklist

-   [x] This pull request links to an issue.
-   [x] This pull request uses the `develop` branch as its base.
-   [x] This pull request subject follows the Conventional Commits standard.
-   [x] This code follows the style guidelines of this project.
-   [x] I have performed a self-review of my code.
-   [ ] I have checked that this code is commented where required.
-   [x] I have provided a detailed enough description of this PR.
-   [x] I have checked if documentation needs updating.

## Documentation

Documentation will definitely need an update, once this is ready for merge. :)


lgsm/functions/workshop_core.sh
```diff
@@ -0,0 +1,318 @@
...
+}
+
+fn_workshop_get_latest_mod_version() {
+	local modid="$1"
+	local serverresp="$(curl -s -d "itemcount=1&publishedfileids[0]=${modid}" "http://api.steampowered.com/ISteamRemoteStorage/GetPublishedFileDetails/v1")"
+	local remupd=
+	if [[ "${serverresp}" =~ \"hcontent_file\":[[:space:]]*([^,]*) ]]; then
+       remupd="${BASH_REMATCH[1]}"
+    fi
+    echo "${remupd}" | tr -d '"'
```
terminalforlife, line 117: Instead of loading up tr(1) and using a subshell (via pipe) for this, it's far more efficient to use BASH's parameter expansion's pattern substitution, as it's already in the language and it's a very small amount of data.

```bash
echo "${remupd//\"/}"
```

Two initial '/' indicate a greedy match, while one is non-greedy.

This feature is available from at least BASH 3.0, including the option to omit the final '/' if no replacement string is provided.

lgsm/functions/workshop_core.sh
```diff
@@ -0,0 +1,318 @@
...
+
+# ## Directory management.
+
+# Create mods files and directories if it doesn't exist.
+fn_create_workshop_dir() {
+	# Create lgsm data modsdir.
+	if [ ! -d "${workshopmodsdldir}" ]; then
+		echo -en "creating LinuxGSM Steam Workshop data directory ${workshopmodsdldir}..."
+		mkdir -p "${workshopmodsdldir}"
+		exitcode=$?
```
terminalforlife, line 257: This is redundant. Sometimes it's useful to specifically catch `$?`, but this is mainly when you're testing for something other than just 0 or non-0, in my experience.

In this case, it's easier and more practical to directly act upon the exit status, rather than separately store the status, check the value, then act based on the value, especially if you're using `[` (inefficient). The `if` statement in shell programming essentially does this for us.

For example:

```bash
if mkdir -p "${workshopmodsdldir}"; then
    ...
else
    ...
fi
```

While this isolated case isn't a huge deal, a lot of redundant code can be a problem, both to readability and performance.

This is core BASH functionality.

lgsm/functions/workshop_core.sh
```diff
@@ -0,0 +1,318 @@
...
+		if [ "${exitcode}" != 0 ]; then
```
terminalforlife, line 258: For future reference, `-eq` is the _numeric_ comparison operator for _equal to_ in BASH, whether in `[` or `[[`. The `=` or `==` is a _string_ comparison operator for _equal to_. Admittedly, the difference isn't often too important, but it's worth knowing and worth using the correct syntax to improve readability and avoid anything unexpected.

See CONDITIONAL EXPRESSIONS in bash(1), and note that the ARITHMETIC EVALUATION section is different, for expanding or evaluating arithmetic in, for example, `((...))`, `$((...))`, `${NAME[...]}`, etc.

lgsm/functions/workshop_core.sh
```diff
@@ -0,0 +1,318 @@
...
+			fn_print_ok_eol_nl
+			fn_script_log_pass "Creating mod install directory ${workshopmodsdir}"
+		fi
+	fi
+}
+
+# Counts how many mods were installed.
+fn_workshop_count_installed() {
+	if [ -f "${workshopmodsdir}" ]; then
+		installedworkshopmodscount=$(ls -l "${workshopmodsdir}" | grep -c ^d)
```
terminalforlife, line 286: It's almost never a good idea to include ls(1) in a script, because it's typically redundant, inefficient, and unreliable. The shell already supports glob filename pattern matching, and counting results is simple.

For example

```bash
Count=0
for File in /path/to/files/*; {
    (( Count++ ))
}

printf '%d\n' $Count
```

Or, more concisely and more efficiently, albeit with far less flexibility:

```bash
Files=(/path/to/files/*)
printf '%d\n' ${#Files[@]}
```

This last one is an awesome feature that often gets overlooked.

These features are available in at least >= BASH 3.0.

I'm not sure why you're looking for '^d', so I apologise if I've missed a nuanced and obscure detail. Granted, BASH can effortlessly do what grep(1) is doing here as well.

For example:

```bash
Count=0
for File in *; {
    if [[ $File == *^d* ]]; then
        (( Count++ ))
    fi
}

printf '%d\n' $Count
```

Provided you're looking for '^d' being in filenames. By using these features, that's another 2 programs and a subshell you'd take out of the equation.

lgsm/functions/workshop_core.sh
```diff
@@ -0,0 +1,318 @@
...
+		echo -e " * Install mods using LinuxGSM first with: ./${selfname} workshop-install"
+		fn_script_log_error "No installed workshop mods or addons were found."
+		core_exit.sh
+	fi
+}
+
+# Builds list of installed Steam Workshop mods.
+fn_workshop_installed_list() {
+	fn_workshop_count_installed
+	for folder in ${workshopmodsdir}/*; do
```
terminalforlife, line 309: Unprotected variable, although there are a few of these. I'm not one of those people who insists on everything being quote-protected, because not everything needs to be, such as integers. However, since it's a directory, one which is presumably changeable by the user or a developer, it's always best to quote-protect it. Unprotected variables like paths can lead to destructive results.

lgsm/functions/workshop_core.sh
```diff
@@ -0,0 +1,318 @@
...
+       remupd="${BASH_REMATCH[1]}"
+    fi
+    echo "${remupd}" | tr -d '"'
+}
+
+fn_workshop_get_name_from_steam() {
+	local modid="$1"
+	local serverresp="$(curl -s -d "itemcount=1&publishedfileids[0]=${modid}" "http://api.steampowered.com/ISteamRemoteStorage/GetPublishedFileDetails/v1")"
+	local title=
+	if [[ "${serverresp}" =~ \"title\":[[:space:]]*([^,]*) ]]; then
```
terminalforlife, line 124: Note that REGEX in BASH, while an awesome feature, is slower than glob pattern matching. You might be thinking that glob pattern matching is limited, so you need REGEX. I used to think the same. However, `[[` uses extended globbing out of the box.

For example:

```bash
if [[ "${serverresp}" == *\"title\":*([[:space:]])(*([^,]))* ]]; then
```

Please test first, in-case I missed something. It's almost 5am. This took much longer than I thought it would.

Refer to `extglob` in bash(1) for more information. It's a useful feature which _usually_ needs to be manually enabled with `shopt -s extglob`, with the exception of `[[`. It's not particularly advanced, but it does cater to most of the common use-cases.

Extended globbing is available in at least >= BASH 3.0, but it being included in `[[` is in >= BASH 4.1.

FliesWithWind:
Thanks for the awesome review! I'll apply your suggestions once I have more time. :)

Dr-Shadow:
@FliesWithWind Do you plan to resume your work on this soon ? I would like to add a specific contribution for DayZ Mods which are currently supported only if we rename files with uppercase characters to lowercase.

terminalforlife:
@Dr-Shadow 

This project is alien to me and I might've misinterpreted your goal, so please take this with a grain of salt.

You can bulk-rename files to lowercase with:

```bash
for File in *; do
    if [[ -f $File ]]; then
        mv "$File" "${File,,}"
    fi
done
```

The `,,` feature of parameter expansion was introduced in BASH 4.0.

That would immediately rename _all_ files in the current working directory to their lowercase counterpart. This is a non-recursive approach, by the way.

However, this should be done _very_ carefully, perhaps with `--no-clobber`, because mv(1) will by default blast away any existing files, should they exist. Another issue with bulk-renaming files in this manner is that you may have conflicts, which you'd have to handle afterwards, such as 'FilE' and 'FILE' being completely different, but both being the same when renamed to all lowercase.

FliesWithWind:
@Dr-Shadow at the moment I have my hands full, but I will try to find some time this month to apply suggestions from @terminalforlife. You are more than welcome to contribute. Arma 3 mod files and folders also need to be in lowercase, so it will likely be the same for DayZ.

PR #4556, opened 2024-04-11
Added optional pre_restart script to be called when monitor restarts …

dualarrow:
These changes allows a script named "pre_restart" to be run (only if it exists) when the monitor command determines a server restart is required.

This is to allow rust updates and oxide updates to be performed when monitor detects a restart is required. Without this, plugins like SmoothRestarter (which can detect a rust and/or oxide update) only stops the server. As rustserver currently stands, the monitor command (as I understand it) only restarts the server and is not able to perform a rust or oxide update.

### DIFF
--- a/lgsm/modules/command_monitor.sh
+++ b/lgsm/modules/command_monitor.sh
@@ -173,6 +173,13 @@ fn_monitor_check_update_source() {
 	fi
 }
 
+fn_pre_restart() {
+	if [ -f "pre_restart" ]; then
+		fn_script_log_info "Running script pre_restart"
+		bash pre_restart
+	fi
+}
+
 fn_monitor_check_session() {
 	fn_print_dots "Checking session: "
 	fn_print_checking_eol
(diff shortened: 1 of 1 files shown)


MicLieg:
Thanks for the PR. That's a great idea!

However using the built in `fixes` feature is probably the better way of achieving this.
See https://dev-docs.linuxgsm.com/getting-started/adding-a-new-game-server#add-any-fixes-to-a-fix-file

Rust already uses some ([pre_start](https://github.com/GameServerManagers/LinuxGSM/blob/master/lgsm/modules/fix.sh#L55)) fixes: https://github.com/GameServerManagers/LinuxGSM/blob/master/lgsm/modules/fix_rust.sh

dualarrow:
I tried that, but it doesnt seem to work.

I added the following to core_modules.sh
```
fix_rust.sh() {
        modulefile="${FUNCNAME[0]}"
        fn_fetch_module
}

fix_rusticraiders.sh() {
        modulefile="${FUNCNAME[0]}"
        fn_fetch_module
}
```
I added the following to fix.sh

`apply_pre_start_fix=(arma3 armar ark av bt bo csgo cmw dst hw ins kf nmrih onset pvr ro rusticraiders rust rw samp sdtd sfc sof2 squad st tf2 terraria ts3 mcb mta unt vh wurm zmr)`

and the fix_rusticraiders.sh is

```
#!/bin/bash
# LinuxGSM fix_rusticraiders.sh module
# Author: Andrew Cumming
# Contributors: http://linuxgsm.com/contrib
# Website: https://linuxgsm.com
# Description: Run some specific tasks before startup

moduleselfname="$(basename "$(readlink -f "${BASH_SOURCE[0]}")")"

echo "SIGN HERE" >./sign.txt
```

The game starts ok, but my script seems to have not executed.
Can you see anything obviously wrong in my example ?

MicLieg:
Currently, each game only has a single fix file containing one or more fixes.

You just need to copy your new code from `fix_rusticraiders.sh` to `fix_rust.sh`. There is no need to create a new fix file, as rust already has one: https://github.com/GameServerManagers/LinuxGSM/blob/master/lgsm/modules/fix_rust.sh

PR #4860, opened 2025-12-01
fix: Set TERM=screen for all non-interactive tmux calls

silasjmatson:
# Description

  When connected to a server via ssh, and the server doesn't have the
  proper terminfo, tmux will fail with 'unsupported terminal' errors.

  These errors are not surfaced by the scripts, e.g. check_status, 
  so they result in either silent failures or don't work at all.

    e.g. `./mcserver details` on my server showed 0% CPU, 0% Memory,
          and STOPPED for the server status when the server was
          actually running. This was because the tmux calls to get
          this info failed.

          The root cause of this was that I was connected to my server
          via kitty, a terminal emulator that my server did not have
          terminfo for.

  This can be resolved via running `infocmp -x xterm-kitty | ssh YOUR-SERVER -- tic -x -`
  and replacing `xterm-kitty` with whatever is appropriate, or by setting
  TERM=xterm-256color in the ssh config, but that is not obvious to do 
  when presented with silent failures.

  This commit sets `TERM=screen` for all non-interactive tmux calls,
  which resolves the issue. Note that the `console` command does not
  set this, as that is an interactive shell, and the user's terminfo
  should be passed.

Pastebin log demonstrating the issue with ghostty terminal: https://pastebin.com/JeAVUfk4

Fixes: #4861

## Type of change

-   [x] Bug fix (a change which fixes an issue).

## Checklist

-   [x] This pull request links to an issue.
-   [x] This pull request uses the `develop` branch as its base.
-   [x] This pull request subject follows the Conventional Commits standard.
-   [x] This code follows the style guidelines of this project.
-   [x] I have performed a self-review of my code.
-   [x] I have checked that this code is commented where required.
-   [x] I have provided a detailed enough description of this PR.
-   [x] I have checked if documentation needs updating.

## Documentation

-   User docs: https://github.com/GameServerManagers/LinuxGSM-Docs
-   Dev docs: https://github.com/GameServerManagers/LinuxGSM-Dev-Docs


dgibbs64:
Really good PR. merging now

Issue #4889, opened 2026-03-27
[Bug]: mcserver on Ubuntu 24.04 still uses OpenJDK 21 while the last Minecraft version (26.1) requires Java 25


Huguiz:
### User story

As a LinuxGSM user on Ubuntu 24.04, I want mcserver to use Java 25 so that current Minecraft can run correctly.

### Game

mcserver

### Linux distro

Ubuntu 24.04

### Command

command: start, command: debug

### Further information

The latest version of Minecraft ([26.1](https://www.minecraft.net/en-us/article/minecraft-java-edition-26-1)) requires Java 25, but the Ubuntu 24 distribution only includes version 21.

This appears related to the earlier Java version mismatch discussed in #4744.

Temporary workaround:
```
apt-get update
apt-get install -y openjdk-25-jre-headless
```

### Relevant log output

```shell
...
Checking Minecraft Dependencies
=================================
2026-03-27T16:19:09.707155734Z checking GitHub ubuntu-24.04.csv ... OK
fetching GitHub [ ubuntu-24.04.csv ] ... OK
bc
binutils
bsdmainutils
bzip2
ca-certificates
cpio
curl
distro-info
file
gzip
hostname
jq
lib32gcc-s1
lib32stdc++6
openjdk version "21.0.10" 2026-01-20
pigz
python3
tmux
unzip
util-linux
uuid-runtime
wget
xz-utils
Skip! Required dependencies already installed.
Accept Minecraft EULA
=================================
You are required to accept the EULA:
https://account.mojang.com/documents/minecraft_eula
By using auto-install you are indicating your agreement to the EULA.
Installing Minecraft Server
=================================
downloading file [ minecraft_server.26.1.jar ]
2026-03-27T16:19:20.126567382Z #=#=#                                                                          
##O#-#                                                                         

##########                                                                14.5%
####################                                                      29.1%
###############################                                           44.0%
##########################################                                58.4%
######################################################                    76.0%
##################################################################        92.1%
######################################################################## 100.0%
downloading file [ minecraft_server.26.1.jar ] ... OK
clearing tmp directory [ /app/lgsm/tmp ] ... OK
Installing Gamedig
% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
Dload  Upload   Total   Spent    Left  Speed
2026-03-27T16:19:20.537490608Z 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[...]

PR #4906, opened 2026-04-26
papermc: add OpenJDK-25 (headless) dependency

WildPenquin:
For a bit of pedantry, use headless. It will save just a little bit of space =)

# Description

Dev builds of papermc will not work on older jre, albeit it's still in alpha/beta, some time in the future it will be the stable build. I have not noticed any issues running the stable build with jre-25.

https://github.com/GameServerManagers/LinuxGSM/issues/4889 (albeit not specific to PaperMC, should I make one?)
https://github.com/GameServerManagers/docker-gameserver/issues/153 (but wrong repo)

## Type of change

- [X] Bug fix (a change which fixes an issue).

## Checklist

- [X] This pull request links to an issue.
- [X] This pull request uses the `develop` branch as its base.
- [X] This pull request subject follows the Conventional Commits standard.
- [X] This code follows the style guidelines of this project.
- [X] I have performed a self-review of my code.
- [X] I have checked that this code is commented where required.
- [X] I have provided a detailed enough description of this PR.
- [X] I have checked if documentation needs updating.

### DIFF
--- a/lgsm/data/ubuntu-24.04.csv
+++ b/lgsm/data/ubuntu-24.04.csv
@@ -82,7 +82,7 @@ onset,libmariadb-dev
 opfor
 pc
 pc2
-pmc,openjdk-21-jre
+pmc,openjdk-25-jre-headless
 pvkii
 pvr,libc++1
 pw


WildPenquin:
Oh also I'm not sure if there are any use cases where a user might require a non-headless JRE (plugins?). If so, it may be safer to just drop the `-headless` part.... the full JRE does not require that much more space anyways.

dgibbs64:
https://docs.papermc.io/misc/java-install/

Do not use headless variants of Java!

There are headless variants of Java which usually have a suffix of -headless in their package name. Those variants miss required dependencies for Paper. Therefore, using them is not recommended.

WildPenquin:
Oh, TIL!

I didn't notice any ill effects so far in this case, but better be safe than sorry.

